PR #64, opened 2023-02-27
feat: chat room

### LINKED ISSUE #39: Chat Room
Chat room for live communication with other members.

SongoMen:
## Proposed changes

This PR adds chat functionality to the backend and frontend, enabling users to send and receive messages in real-time. The backend now uses a Redis pool to publish messages to channels and a single Redis connection to retrieve messages and pass them on to the user's subscription.

When a user enters a chat, they will send a subscription request to the server responsible for handling all messages for that chat. The max message length is 500 and the amount of messages that will be visible on the chat is 300.

## Types of changes

## Checklist

## Further comments

Closes #39

### DIFF
--- a/backend/api/src/api/v1/gql/chat.rs
+++ b/backend/api/src/api/v1/gql/chat.rs
@@ -1,0 +1,80 @@
+use crate::api::v1::gql::error::ResultExt;
+use crate::database::chat_message;
+use crate::pb;
+use prost::Message;
+
+use super::error::{GqlError, Result};
+use super::ext::ContextExt;
+use super::models::chat_message::ChatMessage;
+use async_graphql::{Context, Object};
+use fred::prelude::PubsubInterface;
+use uuid::Uuid;
+
+const MAX_MESSAGE_LENGTH: usize = 500;
+
+#[derive(Default)]
+pub struct ChatMutation;
+
+#[Object]
+impl ChatMutation {
+    // Send message in chat. You need to be logged in for that.
+    async fn send_message<'ctx>(
+        &self,
+        ctx: &Context<'_>,
+        #[graphql(desc = "ID of chat room where the message will be send.")] channel_id: Uuid,
+        #[graphql(desc = "Message content that will be published.")] content: String,
+    ) -> Result<ChatMessage> {
+        let global = ctx.get_global();
+        let request_context = ctx.get_session();
+
+        if content.len() > MAX_MESSAGE_LENGTH {
+            return Err(GqlError::InvalidInput.with_message("Message too long"));
+        }
+
+        // TODO: check if user is banned from chat
+        let (session, _) = request_context
+            .get_session(global)
+            .await?
+            .ok_or_else(|| GqlError::Unauthorized.with_message("You need to be logged in"))?;
+
+        // TODO: Check if the user is allowed to send messages in this chat
+        let channel = global
+            .user_by_id_loader
+            .load_one(channel_id)
+            .await
+            .map_err_gql("Failed to fetch channel")?
+            .ok_or_else(|| GqlError::InvalidInput.with_message("Channel not found"))?;
+
+        let chat_message = sqlx::query_as!(
+            chat_message::Model,
+            "INSERT INTO chat_messages (channel_id, author_id, content) VALUES ($1, $2, $3) RETURNING *",
+            channel.id,
+            session.user_id,
+            content,
+        ).fetch_one(&*global.db).await.map_err_gql("Failed to insert chat message")?;
+
+        match global
+            .redis
+            .publish(
+                format!("user:{}:chat:messages", channel.id),
+                pb::scuffle::events::ChatMessage {
+                    id: chat_message.id.to_string(),
+                    channel_id: chat_message.channel_id.to_string(),
+                    author_id: chat_message.author_id.to_string(),
+                    content: chat_message.content.clone(),
+                    created_at: chat_message.created_at.timestamp(),
+                }
+                .encode_to_vec()
+                .as_slice(),
+            )
+            .await
+        {
+            Ok(()) => {}
+            Err(_) => {
+                return Err(GqlError::InternalServerError.with_message("Failed to publish message"));
+            }
+        };
+
+        Ok(chat_message.into())
+    }
+}
--- a/backend/api/src/api/v1/gql/mod.rs
+++ b/backend/api/src/api/v1/gql/mod.rs
@@ -13,6 +13,7 @@ use self::{
 };
 
 pub mod auth;
+pub mod chat;
 pub mod error;
 pub mod ext;
 pub mod handlers;
@@ -31,6 +32,7 @@ pub struct Query {
 /// The root mutation type which contains root level fields.
 pub struct Mutation {
     auth: auth::AuthMutation,
+    chat: chat::ChatMutation,
 }
 
 #[ComplexObject]
--- a/backend/api/src/api/v1/gql/subscription/chat.rs
+++ b/backend/api/src/api/v1/gql/subscription/chat.rs
@@ -1,0 +1,70 @@
+use async_graphql::{Context, Subscription};
+use async_stream::stream;
+use chrono::{TimeZone, Utc};
+use futures_util::Stream;
+use prost::Message;
+use uuid::Uuid;
+
+use crate::{
+    api::v1::gql::{
+        error::{GqlError, Result, ResultExt},
+        ext::ContextExt,
+        models::chat_message::{ChatMessage, MessageType},
+    },
+    pb,
+};
+
+#[derive(Default)]
+pub struct ChatSubscription;
+
+#[Subscription]
+impl ChatSubscription {
+    // Listen to new messages in chat.
+    pub async fn chat_messages<'ctx>(
+        &self,
+        ctx: &'ctx Context<'_>,
+        #[graphql(desc = "Chat to subscribe to.")] channel_id: Uuid,
+    ) -> Result<impl Stream<Item = Result<ChatMessage>> + 'ctx> {
+        let global = ctx.get_global();
+
+        let welcome_message = ChatMessage {
+            id: Uuid::nil(),
+            author_id: Uuid::nil(),
+            channel_id,
+            content: "Welcome to the chat!".to_string(),
+            created_at: chrono::Utc::now().into(),
+            r#type: MessageType::Welcome,
+        };
+
+        // TODO: check if user is allowed to read this chat
+        let channel = global
+            .user_by_id_loader
+            .load_one(channel_id)
+            .await
+            .map_err_gql("failed to fetch user")?
+            .ok_or(GqlError::NotFound.with_message("user not found"))?;
+        let mut message_stream = global
+            .subscription_manager
+            .subscribe(format!("user:{}:chat:messages", channel.id))
+            .await
+            .map_err_gql("failed to subscribe to chat messages")?;
+
+        Ok(stream! {
+            yield Ok(welcome_message);
+            while let Ok(message) = message_stream.recv().await {
+                let event = pb::scuffle::events::ChatMessage::decode(
+                    message.as_bytes().map_err_gql("invalid redis value type")?
+                ).map_err_gql("failed to decode chat message")?;
+
+                yield Ok(ChatMessage {
+                    id: Uuid::parse_str(&event.id).map_err_gql("failed to parse chat message id")?,
+                    author_id: Uuid::parse_str(&event.author_id).map_err_gql("failed to parse chat message author id")?,
+                    channel_id: Uuid::parse_str(&event.channel_id).map_err_gql("failed to parse chat message channel id")?,
+                    content: event.content,
+                    created_at: Utc.timestamp_opt(event.created_at, 0).single().map_err_gql("failed to parse chat message created at")?.into(),
+                    r#type: MessageType::User,
+                });
+            }
+        })
+    }
+}
--- a/backend/api/src/api/v1/gql/subscription/user.rs
+++ b/backend/api/src/api/v1/gql/subscription/user.rs
@@ -1,5 +1,4 @@
 use async_graphql::{Context, SimpleObject, Subscription};
-use fred::types::RedisValue;
 use futures_util::Stream;
 use prost::Message;
 use uuid::Uuid;
@@ -36,7 +35,7 @@ impl UserSubscription {
 
         let mut subscription = global
             .subscription_manager
-            .subscribe(format!("user.{}.display_name", user_id))
+            .subscribe(format!("user:{}:display_name", user_id))
             .await
             .map_err_gql("failed to subscribe to user display name")?;
 
@@ -46,16 +45,10 @@ impl UserSubscription {
                 username: user.username.clone(),
             });
 
-            while let Ok(value) = subscription.recv().await {
-                let value = match value {
-                    RedisValue::String(val) => val,
-                    _ => {
-                        yield Err(GqlError::InternalServerError.with_message("invalid redis value type"));
-                        break;
-                    },
-                }.into_inner();
-
-                let event = pb::scuffle::types::api::UserDisplayName::decode(value).map_err_gql("failed to decode user display name")?;
+            while let Ok(message) = subscription.recv().await {
+                let event = pb::scuffle::events::UserDisplayName::decode(
+                    message.as_bytes().map_err_gql("invalid redis value")?
+                ).map_err_gql("failed to decode user display name")?;
 
                 if let Some(username) = event.username {
                     user.username = username;
--- a/backend/api/src/tests/api/v1/gql/chat.rs
+++ b/backend/api/src/tests/api/v1/gql/chat.rs
@@ -1,0 +1,358 @@
+use crate::{
+    api::v1::gql::ext::RequestExt,
+    database::{chat_message, session, user},
+    pb,
+};
+use async_graphql::{Name, Request, Variables};
+use chrono::Utc;
+use common::prelude::FutureTimeout;
+use prost::Message;
+use serial_test::serial;
+use std::sync::Arc;
+use uuid::Uuid;
+
+use crate::{
+    api::v1::gql::{request_context::RequestContext, schema},
+    tests::global::mock_global_state,
+};
+
+#[tokio::test]
+#[serial]
+async fn test_serial_send_message_not_logged_in() {
+    let (global, _handler) = mock_global_state(Default::default()).await;
+    let schema = schema();
+
+    let query = r#"
+        mutation SendChatMessage($channelId: UUID!, $content: String!) {
+            chat {
+                sendMessage(channelId: $channelId, content: $content) {
+                    id
+                }
+            }
+        }
+    "#;
+
+    let ctx = Arc::new(RequestContext::default());
+
+    let mut variables = Variables::default();
+    variables.insert(
+        Name::new("channelId"),
+        async_graphql::Value::String(Uuid::new_v4().to_string()),
+    );
+
+    variables.insert(
+        Name::new("content"),
+        async_graphql::Value::String("message".to_owned()),
+    );
+
+    let res = schema
+        .execute(
+            Request::from(query)
+                .variables(variables)
+                .provide_global(global.clone())
+                .provide_context(ctx),
+        )
+        .await;
+
+    println!("{:?}", res);
+
+    assert_eq!(res.errors.len(), 1);
+    let json = res.data.into_json();
+    assert!(json.is_ok());
+    assert_eq!(json.unwrap(), serde_json::json!(null));
+    assert_eq!(
+        res.errors[0].message,
+        "Unauthorized: You need to be logged in"
+    );
+}
+
+#[tokio::test]
+#[serial]
+async fn test_serial_send_message_chat_not_found() {
+    let (global, _handler) = mock_global_state(Default::default()).await;
+    let schema = schema();
+    let query = r#"
+        mutation SendChatMessage($channelId: UUID!, $content: String!) {
+            chat {
+                sendMessage(channelId: $channelId, content: $content) {
+                    id
+                }
+            }
+        }
+    "#;
+
+    sqlx::query!("DELETE FROM sessions")
+        .execute(&*global.db)
+        .await
+        .unwrap();
+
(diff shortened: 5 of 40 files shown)


TroyKomodo:
This will have to go in after #63

TroyKomodo:
I think, initially we should focus on doing chat via the GQL subscription model.
Where users can subscript to a chat stream and GQL subscriptions will handle the emit events for us. 
We can add support for a raw websocket version as well as an SSE version later when we do a full Rest API. however initially we should just stick with a simple GQL subscription.

Secondly, this design makes use of tokio's broadcast channels. This is great for a single instance application but if we scale this app our horizontally, instances of the application need to be able to receive the chat messages which are sent to other nodes in the cluster. 

The only realistic way to go about the 2nd point is to have an event emitting system. 

Currently we use postgres for a database however it does support a pubsub model and can be used as a fanout event emitter. 
Redis is also an option.
Both have pros and cons.

For postgres, as a pubsub I am not entirely sure of the performance around that. I know it is very nice because we can have events (such as updates, inserts, deletes) trigger pubsub so a simple query `INSERT INTO chat_messages (x, y, z) ($1, $2, $3)` will cause a pubsub emit to some custom defined endpoint based on the data inserted into the table. However I am not too sure about the relative performance of that. 

Redis, redis is very very battle tested and their pubsub is currently what powers 7tv event sub and we have never had any issues with the scalability of that (when it comes to the actual event sub, not the long lived connections) The disadvantage to redis is we will have to emit an event (on insert) manually. Also redis is unreliable (it does not care if anyone is listening to the event) in this case its fine because chat messages can be dropped and its not very important. 

Both are super strong cases. If postgres is performant enough I would opt to go for that. Perhaps as part of this ticket you could look at the relative performance of how postgres scales with subscriptions and triggers on inserts / updates / deletes ect.

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,80 @@
...
+        #[graphql(desc = "Chat to subscribe to.")] id: String,
+    ) -> impl Stream<Item = Message> {
+        let global = ctx
+            .data::<Arc<GlobalState>>()
+            .expect("failed to get global state")
+            .clone();
+
+        let redis_conn = global.redis_sub_pool.next();
+        let mut message_stream = redis_conn.on_message();
+        let _ = redis_conn.subscribe(id).await;
```
TroyKomodo, line 69: I believe fred has a subscription helper, with this impl each subscription will require its own redis connection when in reality you can just resubscribe to more topics. Rather dont use a redis pool for subs but instead a single connection.

TroyKomodo: Fred has some helpers which make this super easy

backend/api/src/config.rs
```diff
@@ -16,6 +16,9 @@ pub struct AppConfig {
     /// The database URL to use
     pub database_url: String,
 
+    /// The redis URL to use
+    pub redis_url: String,
```
TroyKomodo, line 20: We also make use of Redis Sentinel.

Redis Sentinel is a way to make a HA Redis Cluster. Since the master switches to a slave if it becomes unresponsive or is shutting down.

So perhaps have an option to enable sentinel and instead convert this to 
```
redis_urls: Vec<String>,
redis_username: String,
redis_password: String,
redis_sentinel: bool,
```

backend/api/src/config.rs
```diff
@@ -16,6 +16,18 @@ pub struct AppConfig {
     /// The database URL to use
     pub database_url: String,
 
+    /// The Redis URLs to use
+    pub redis_urls: Vec<(String, u16)>,
```
TroyKomodo, line 20: when it is a tuple how does this get parsed from an env variable?

TroyKomodo: or a file?

backend/api/src/tests/global/mod.rs
```diff
@@ -20,5 +23,43 @@ pub async fn mock_global_state(config: AppConfig) -> (Arc<GlobalState>, Handler)
...
+            },
+            ..Default::default()
+        }
+    } else {
+        let (host, port) = &config.redis_urls[0];
+        RedisConfig {
+            server: ServerConfig::new_centralized(host.clone(), *port),
+            ..Default::default()
+        }
+    };
```
TroyKomodo, line 44: Perhaps we abstract this logic to a method?

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let user = sqlx::query!("SELECT username FROM users WHERE id = $1", user_id,)
+            .map(|row| user::Model {
+                username: row.username,
+                ..Default::default()
+            })
+            .fetch_optional(&*global.db)
+            .await
+            .map_err(|_| GqlError::InvalidInput.with_message("Failed to fetch user"))?
+            .ok_or_else(|| GqlError::InvalidInput.with_message("User not found"))?;
```
TroyKomodo, line 58: We should use the dataloaders

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let _chat = sqlx::query!("SELECT id FROM chat_rooms WHERE id = $1", chat_id,)
+            .map(|row| chat_room::Model {
+                id: row.id,
+                ..Default::default()
+            })
+            .fetch_optional(&*global.db)
+            .await
+            .map_err(|_| GqlError::InvalidInput.with_message("Failed to fetch chat room"))?
+            .ok_or_else(|| GqlError::InvalidInput.with_message("Chat not found"))?;
```
TroyKomodo, line 68: Should make a new dataloader for this.

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let message_to_send = serde_json::json!({ "chat_id": chat_id, "username": user.username, "content": content, "message_type": "message" }).to_string();
```
TroyKomodo, line 70: Perhaps convert this to an actual struct rather than a json inline

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let mut message_stream = global.redis_sub_client.on_message();
+        let _ = global
+            .redis_sub_client
+            .subscribe("chat:".to_string() + &chat_id.to_string())
+            .await;
+
+        stream! {
+            loop {
+                while let Some((channel, message)) = message_stream.next().await {
```
TroyKomodo, line 120: Double loops?

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,131 @@
...
+        let global = ctx
+            .data::<Arc<GlobalState>>()
+            .expect("failed to get global state")
+            .clone();
+
+        let mut message_stream = global.redis_sub_client.on_message();
+        let _ = global
+            .redis_sub_client
+            .subscribe("chat:".to_string() + &chat_id.to_string())
+            .await;
```
TroyKomodo, line 116: Not sure of this approach, it seems that you have a single global broadcast channel?
Perhaps we need to change how this works.

backend/api/src/api/v1/gql/mod.rs
```diff
@@ -77,6 +79,17 @@ impl Subscription {
...
+    async fn new_message(
+        &self,
+        ctx: &Context<'_>,
+        chat_id: i64,
+    ) -> Result<impl Stream<Item = models::message::Message>> {
+        Ok(chat::ChatSubscription
+            .new_message(ctx, chat_id)
+            .await
+            .unwrap())
+    }
```
TroyKomodo, line 92: It seems you have this method duplicated why not add it the same way as a field on the subscription object?

frontend/website/tsconfig.json
```diff
@@ -8,7 +8,13 @@
...
 		"skipLibCheck": true,
 		"sourceMap": true,
-		"strict": true
+		"strict": true,
+		"paths": {
+			"$lib": ["./src/lib"],
+			"$lib/*": ["./src/lib/*"],
+			"$components": ["./src/components"],
+			"$components/*": ["./src/components/*"]
+		}
```
TroyKomodo, line 17: I dont think this works. See the comment under.

backend/api/Cargo.toml
```diff
@@ -26,6 +26,7 @@ serde = { version = "1.0.152", features = ["derive"] }
 hyper = { version = "0.14.24", features = ["full"] }
 common = { path = "../../common" }
 sqlx = { version = "0.6.2", features = ["postgres", "runtime-tokio-rustls", "json", "chrono"] }
+fred = { version = "5.2.0", features=["subscriber-client"]}
```
TroyKomodo, line 29: you likely need to switch to `rustls` from `openssl`

backend/api/src/api/v1/gql/chat.rs
SongoMen: Do you have any idea?

backend/api/src/api/v1/gql/mod.rs
SongoMen: I had to use `MergedSubscription` I couldn't get it to work another way

TroyKomodo:
@SongoMen Can you rebase with main?

frontend/website/src/components/chatroom.svelte
```diff
@@ -100,10 +159,17 @@
 		<span class="chat-title">Chat</span>
 	</div>
 	<div class="messages" bind:this={messagesEl}>
-		{#each messages as message (message.svelte_id)}
-			<div class="message" class:odd={message.svelte_id % 2}>
-				<span class="message-sender" style={`color: ${message.color}`}>{message.sender}:</span>
-				<span class="message-text">{message.text}</span>
+		{#each messages as message}
+			<div class="message">
```
TroyKomodo, line 163: No even odd messages?

frontend/website/src/components/chatroom.svelte
```diff
@@ -100,10 +159,17 @@
...
+				{#if message.messageType === "message"}
+					<span class="message-sender" style={`color: ${message.metadata.color}`}
```
TroyKomodo, line 165: Where does the color come from?
At the moment user's dont have a color in the database.

frontend/website/tsconfig.json
TroyKomodo: interesting

frontend/website/svelte.config.js
```diff
@@ -27,6 +27,10 @@ const config = {
 				config.include.push("../wasm.d.ts");
 			},
 		},
+		alias: {
+			"components/*": "src/components/*",
+			"lib/*": "src/lib/*",
+		},
```
TroyKomodo, line 33: Could you update everywhere where we use relative paths.

TroyKomodo: Perhaps also add an alias for `$` to be the root

schema.graphql
```diff
@@ -78,7 +95,11 @@ type Session {
 	userId: Int!
 }
 
+"""
+The root subscription type which contains root level fields.
+"""
```
TroyKomodo, line 100: Where is this comment from?

schema.graphql
```diff
@@ -33,13 +33,30 @@ type AuthMutation {
 	): Session!
 }
 
+type ChatMutation {
+	sendMessage(chatId: Int!, content: String!): String!
```
TroyKomodo, line 37: As mentioned before this should return a Message object

backend/api/src/tests/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    "#;
+
+    let ctx = Arc::new(RequestContext::default());
+
+    let res = schema
+        .execute(Request::from(query).data(global.clone()).data(ctx))
+        .await;
+
+    assert_eq!(res.errors.len(), 1);
```
TroyKomodo, line 53: Can we check the error contents, please refer to other tests for how this is done

backend/api/src/tests/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    let ctx = Arc::new(RequestContext::new(false));
+    ctx.set_session(Some(session.to_owned()));
+
+    let res = schema
+        .execute(Request::from(query).data(global.clone()).data(ctx))
+        .await;
+    assert_eq!(res.errors.len(), 1);
+    let json = res.data.into_json();
+    assert!(json.is_ok());
```
TroyKomodo, line 88: Can we check the response value?

backend/api/src/tests/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    let ctx = Arc::new(RequestContext::default());
+
+    let res = schema
+        .execute(Request::from(query).data(global.clone()).data(ctx))
+        .await;
+
+    assert_eq!(res.errors.len(), 1);
+    let json = res.data.into_json();
+    assert!(json.is_ok());
```
TroyKomodo, line 55: Can we check the response value

backend/api/src/tests/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,153 @@
...
+    .await
+    .unwrap();
+
+    let ctx = Arc::new(RequestContext::new(false));
+    ctx.set_session(Some(session.to_owned()));
+
+    let res = schema
+        .execute(Request::from(query).data(global.clone()).data(ctx))
+        .await;
+    assert_eq!(res.errors.len(), 1);
```
TroyKomodo, line 86: Check error content

backend/api/src/tests/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    let ctx = Arc::new(RequestContext::new(true));
+    ctx.set_session(Some(session.to_owned()));
+
+    let res = schema
+        .execute(Request::from(query).data(global.clone()).data(ctx))
+        .await;
+    assert_eq!(res.errors.len(), 0);
+    let json = res.data.into_json();
+    assert!(json.is_ok());
```
TroyKomodo, line 110: Same as above

backend/api/src/tests/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,153 @@
...
+            Request::from(query)
+                .variables(variables)
+                .data(global.clone())
+                .data(ctx),
+        )
+        .await;
+
+    assert_eq!(res.errors.len(), 1);
+    let json = res.data.into_json();
+    assert!(json.is_ok());
```
TroyKomodo, line 152: Same as above

backend/api/src/tests/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,153 @@
...
+use common::types::{session, user};
+use std::sync::Arc;
+
+use crate::{
+    api::v1::gql::{request_context::RequestContext, schema},
+    tests::global::mock_global_state,
+};
+
+#[tokio::test]
+async fn test_send_message() {
```
TroyKomodo, line 12: This test is really long, perhaps we can split it into multiple tests?

TroyKomodo (changes requested):
very good start!

frontend/website/src/components/chatroom.svelte
SongoMen: Its right now hardcoded on backend to teal

schema.graphql
SongoMen: `v1/gql/mod.rs` file

frontend/website/src/components/chatroom.svelte
SongoMen: I thought different style for odd/even messages looked worse? I can add it back if you want, maybe with different background color

frontend/website/src/components/chatroom.svelte
TroyKomodo: ok thats fine

schema.graphql
TroyKomodo: Can it be removed since its not accurate

TroyKomodo:
Thanks for rerunning codecov 
image
code cov reports that the chat subscription event is never hit, could we add a unit test for this subscription event?

As well as the get_redis_config on the AppConfig struct
image

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,143 @@
...
+            chat_id,
+            user_id,
+            content,
+            Utc::now()
+        )
+        .fetch_one(&*global.db)
+        .await
+        .map_err_gql("Failed to send message")?;
+
+        let message_to_send = Message {
```
TroyKomodo, line 69: Perhaps , add a `impl From<chat_message::Model>` onto this struct

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,143 @@
...
+                format!("chat:{}", chat_id),
+                serde_json::json!(message_to_send).to_string(),
+            )
+            .await
+        {
+            Ok(_) => {}
+            Err(_) => {
+                return Err(GqlError::InternalServerError.with_message("Failed to publish message"));
+            }
+        };
```
TroyKomodo, line 90: Perhaps something like this (untested + unformatted)

```suggestion
        if let Err(err) = global.redis_pool.publish::<String, String, String>(format!("chat:{}", chat_id), serde_json::json!(message_to_send).to_string()).await {
            tracing::error!("failed to publish redis message: {}", err)
            return Err(GqlError::InternalServerError.with_message("Failed to publish message"));
        }
```

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        let _chat = match global.chat_room_by_id_loader.load_one(chat_id).await {
+            Ok(Some(chat)) => chat,
+            Ok(None) => {
+                return Err(GqlError::InvalidInput.with_message("Chat not found"));
+            }
+            Err(_) => {
+                return Err(GqlError::InternalServerError.with_message("Failed to fetch chat"));
+            }
+        };
```
TroyKomodo, line 123: Perhaps something like this? (untested + unformatted)

```suggestion
        match global.chat_room_by_id_loader.load_one(chat_id).await {
            Ok(Some(_)) => (),
            Ok(None) => {
                return Err(GqlError::InvalidInput.with_message("Chat not found"));
            }
            Err(err) => {
                tracing::error!("failed to query chat room: {}", err);
                return Err(GqlError::InternalServerError.with_message("Failed to fetch chat"));
            }
        };
```

frontend/website/src/components/chatroom.svelte
```diff
@@ -1,40 +1,76 @@
...
+		return {
+			authorId: "",
+			// We cast to User because this is a system message and we don't care about the author or the channel
+			channel: {} as User,
+			author: {} as User,
+			channelId: channelId,
+			createdAt: "",
+			id: "",
+			type: MessageType.System,
+			content: content,
```
treuks, line 34: leaving it as empty strings and stuff feels odd, could it be `null` instead? feel like that would be nicer

backend/api/src/api/v1/gql/chat.rs
```diff
@@ -0,0 +1,80 @@
...
+            .map_err_gql("Failed to fetch channel")?
+            .ok_or_else(|| GqlError::InvalidInput.with_message("Channel not found"))?;
+
+        let chat_message = sqlx::query_as!(
+            chat_message::Model,
+            "INSERT INTO chat_messages (channel_id, author_id, content) VALUES ($1, $2, $3) RETURNING *",
+            channel.id,
+            session.user_id,
+            content,
+        ).fetch_one(&*global.db).await.map_err_gql("Failed to insert chat message")?;
```
treuks, line 54: works for now, but we will definitely have to switch off from using a traditional database for this in the future. id suggest adding a todo suggesting for us to switch to ClickHouse or something else that is like that in the future

TroyKomodo: how do u mean a traditional database?

backend/api/src/api/v1/gql/subscription/chat.rs
```diff
@@ -0,0 +1,70 @@
...
+        let global = ctx.get_global();
+
+        let welcome_message = ChatMessage {
+            id: Uuid::nil(),
+            author_id: Uuid::nil(),
+            channel_id,
+            content: "Welcome to the chat!".to_string(),
+            created_at: chrono::Utc::now().into(),
+            r#type: MessageType::Welcome,
+        };
```
lennartkloock, line 37: Do we really need to send a welcome message?
Can't we just show the message client-side as soon as the user is connected to the chat?

backend/api/src/api/v1/gql/subscription/user.rs
```diff
@@ -36,7 +35,7 @@ impl UserSubscription {
 
         let mut subscription = global
             .subscription_manager
-            .subscribe(format!("user.{}.display_name", user_id))
+            .subscribe(format!("user:{}:display_name", user_id))
```
lennartkloock, line 38: Why did you switch from using `.` to `:`?
Redis suggests to use dots: https://redis.io/docs/manual/pubsub/#pattern-matching-subscriptions
It probably doesn't matter though.

TroyKomodo: this isnt the pattern based subscription its just a regular one, and in redis namespaces are generally denoted by colons `:`
https://redis.com/blog/5-key-takeaways-for-developing-with-redis/

backend/api/src/api/v1/gql/subscription/chat.rs
lennartkloock: Oh right, didn't think about that

lennartkloock (approved):

backend/api/src/api/v1/gql/chat.rs
TroyKomodo: Spoke in discord: Maybe when we notice performance degradation from chat messages in cockroachdb

treuks (approved):

TroyKomodo (approved):

PR #67, opened 2023-03-05
Dataloaders

TroyKomodo:
## Proposed changes

Dataloaders allow us to batch queries across requests which in turn make it very optimized at scale. Using the `WHERE id IN ANY($1)` SQL feature allows us to combine many queries into a single query which makes it very efficient at high concurrency requests.

We should also investigate how we can take advantage of redis caching to speed up results. The main issue with redis caching is how do we uncache stale data. Results need to be fresh always, that is a MUST there is no way we can serve stale data. So redis caching becomes a complex mess because invalidating cached data is very hard.

## Types of changes

## Checklist

- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/backend/api/src/api/v1/gql/auth.rs
+++ b/backend/api/src/api/v1/gql/auth.rs
@@ -48,23 +48,16 @@ impl AuthMutation {
                 .with_field(vec!["captchaToken"]));
         }
 
-        let user = sqlx::query!(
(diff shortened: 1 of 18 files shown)


chipsTM (approved):
Looks good from my end

4w3Official (approved):
LGTM

Issue #87, opened 2023-05-22
API


TroyKomodo:
API Lacks some functionality
- [ ] Handle resume streams correctly
- [x] Add live updates with pubsub events and GQL subscriptions (#90)
- [ ] Add ability to stop live streams

TroyKomodo:
Closed because this issue are too general, and also no longer valid.

PR #90, opened 2023-05-28
feat: gql pubsub event subscriptions

lennartkloock:
This addresses the second point in #87 

## Proposed changes

This adds support for GQL pubsub event subscriptions for various events.

## Types of changes

## Checklist

### DIFF
--- a/backend/api/src/subscription.rs
+++ b/backend/api/src/subscription.rs
@@ -1,0 +1,161 @@
+use std::{
+    collections::HashMap,
+    ops::{Deref, DerefMut},
+};
+
+use anyhow::Result;
+use common::context::Context;
+use fred::{clients::SubscriberClient, prelude::PubsubInterface, types::RedisValue};
+use tokio::{
+    select,
+    sync::{broadcast, mpsc, oneshot, Mutex},
+};
+
+#[derive(Debug)]
+enum Event {
+    Subscribe {
+        topic: String,
+        tx: oneshot::Sender<broadcast::Receiver<RedisValue>>,
+    },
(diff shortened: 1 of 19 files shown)


lennartkloock (commented):
lgtm

lennartkloock (commented):
still lgtm

backend/api/src/subscription.rs
```diff
@@ -0,0 +1,161 @@
...
+        let mut events_rx = self.events_rx.lock().await;
+
+        let mut messages = redis.on_message();
+
+        loop {
+            select! {
+                event = events_rx.recv() => {
+                    match event.unwrap() {
+                        Event::Subscribe { topic, tx } => {
+                            let topic = topic.to_lowercase();
```
treuks, line 78: is there no better way to do this? why mutate?

SputNikPlop (approved):
LGTM just take a look what @treuks mentioned

backend/api/src/subscription.rs
TroyKomodo: What do u mean?

treuks: image
(from the discord conversation from context). essentially, i think it would be nice if we had a way to parse a string into a specific channel class which would absolutely concretely ensure the channel was correct so it would not have to be rechecked or reparsed into any other class again

treuks: image
alright this makes sense, resolving this now

treuks (approved):

TroyKomodo (approved):

PR #136, opened 2023-10-16
fix: unsound cache upcast

TroyKomodo:
## Proposed changes

Removes the unsafe code around cache upcasting by introducing traited functions `as_ref` and `as_mut` on the AutoImplCacheRef trait. Also no longer requires the `unsafe` attribute.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

## Further comments

If this is a relatively large or complex change, you may want to start the discussion by explaining why you chose the solution you did, what alternatives you considered, etc.

### DIFF
--- a/common/src/dataloader/cache.rs
+++ b/common/src/dataloader/cache.rs
@@ -1,4 +1,7 @@
-use std::collections::hash_map::RandomState;
+use std::{
+    collections::hash_map::RandomState,
+    ops::{Deref, DerefMut},
+};
 
 use crate::dataloader::LoaderOutput;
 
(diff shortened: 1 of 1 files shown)


common/src/dataloader/cache.rs
```diff
@@ -16,62 +19,70 @@ pub trait Cache<L: Loader<S>, S = RandomState> {
...
-#[inline(always)]
-#[allow(clippy::mut_from_ref)]
-#[allow(invalid_reference_casting)]
-fn upcast<T: AutoImplCacheRef<L, S>, L: Loader<S>, S>(cache: &T) -> impl Cache<L, S> {
-    // Safety:
-    // This is safe because the trait `AutoImplCacheRef` is marked as unsafe and therefore the implementor must ensure that the
-    // Cache is safe for concurrent access. This is used to implement cache for &T where T is a Cache.
-    // The issue is we need to upcast the reference to a mutable reference, even though the implementor only ever requires a reference.
-    // This is not safe unless T has some kind of interior mutability.
-    unsafe { &mut *(cache as *const T as *mut T) }
```
TroyKomodo, line 38: There was actually nothing wrong with the way we used this. The issue would have been
```rs
struct CImpl {
   cache: Mutex<i32>
}

impl Cache {
  fn clear(&mut self) {
      *self.cache.get_mut() = 0;
  }
   ...
}
```

If the caller implemented the unsafe trait incorrectly it would be UB; now you might say well that is the point of the unsafe part. and you would be right, but i might as well catch it before they fuck up.

lennartkloock (approved):
My review is pretty useless here because I never wrote any unsafe and don't really know what's going on here. I don't even understand what this was for in the first place.

lgtm

TroyKomodo:
> My review is pretty useless here because I never wrote any unsafe and don't really know what's going on here. I don't even understand what this was for in the first place.
> 
> lgtm

It was so that we can allow for a cache to be a reference, rather than having an owned value. An example of this might be

```rs
let cache = SharedCache::new()

loader.load("key", &cache);
loader.load("key", &cache);
```

Rather than
```rs
let cache = Arc::new(SharedCache::new())

loader.load("key", cache.clone());
loader.load("key", cache.clone());
```

lennartkloock:
> It was so that we can allow for a cache to be a reference, rather than having an owned value. An example of this might be

I see.

Always good to remove unsafe code.

PR #149, opened 2023-11-12
feat: remove binary impl for edge, transcoder, ingest

TroyKomodo:
## Proposed changes

The reasoning behind this is because ideally we should provide a simple way for users to launch all services under a single binary. This change refactors how global state works. Allowing us to change the global state as a trait. This also has the effect of removing code duplication.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, you may want to start the discussion by explaining why you chose the solution you did, what alternatives you considered, etc.


common/src/global.rs
```diff
@@ -0,0 +1,34 @@
...
+    fn nats(&self) -> &async_nats::Client;
+    fn jetstream(&self) -> &async_nats::jetstream::Context;
+}
+
+pub trait GlobalDb {
+    fn db(&self) -> &Arc<sqlx::PgPool>;
+}
+
+pub trait GlobalRedis {
+    // fn redis(&self) -> &Arc<redis::aio::MultiplexedConnection>;
```
lennartkloock, line 33: Why is this commented out?

lennartkloock (approved):

lennartkloock:
Do you plan to use the binary helper for the platform api as well?

TroyKomodo:
@lennartkloock yeah likely

lennartkloock:
> @lennartkloock yeah likely

In this pr?

4w3Official (approved):

binary-helper/src/config.rs
```diff
@@ -0,0 +1,82 @@
...
+        AppConfig {
+            name: Self::APP_NAME.to_owned(),
+            config_file: Some("config".to_owned()),
+            logging: Default::default(),
+            grpc: Default::default(),
+            database: Default::default(),
+            nats: Default::default(),
+            extra: Self::default(),
+        }
+    }
```
lennartkloock, line 47: Why is this a function on the `ConfigExtension` trait instead of an `Default` impl for `AppConfig`

platform/api/src/main.rs
```diff
@@ -1,110 +1,221 @@
...
+        &self.config.extra.jwt
+    }
+}
 
-    // Sets the similarity threshold for trigram matching to 0.1
-    // Default is 0.3
-    // https://www.cockroachlabs.com/docs/stable/trigram-indexes#comparisons
-    sqlx::query("SET pg_trgm.similarity_threshold = 0.1")
-        .execute(&*db)
-        .await?;
```
lennartkloock, line 76: Where did this go?

binary-helper/src/config.rs
TroyKomodo: Rust has namespace rules about trait implementation. AppConfig is in a different crate so i cannot implement traits on that struct.

platform/api/src/main.rs
TroyKomodo: I was wondering if there is a better way to do this? Perhaps changing the dB configuration so it persists rather than executing it every startup. If the dB restarted the connections would restart and the setting would be lost.

binary-helper/src/config.rs
lennartkloock: Isn't AppConfig in this same file?

platform/api/src/main.rs
lennartkloock: Maybe we can include this in a migration?

binary-helper/src/config.rs
TroyKomodo: No, only an alias.

lennartkloock (approved):
lgtm

platform/api/src/main.rs
TroyKomodo: Yeah, if it can be done there or somewhere else.

PR #172, opened 2023-12-30
feat(platform): add file upload endpoint

TroyKomodo:
## Proposed changes

Adds a generic file upload rest endpoint.

The idea is, you will query GQL and the GQL API will return a token you can use to upload a file to. (Something like a singed URL, except its just a token you put into the Authentication Header). We then query (delete) the token in the database to validate that the token has not already been used (deleted). Once we have everything we need, the database will tell us how we need to handle the uploaded file. Right now all we will support is image uploads for Profile pictures, or banners or whatever. But in future we may allow any file upload and it will all use the same endpoint which makes things nice and simple.

## Types of changes

## Checklist

- [ ] Lint and unit tests pass locally with my changes

## Further comments
.-.


lennartkloock (approved):
lgtm

PR #186, opened 2024-01-12
feat(website): profile pictures

lennartkloock:
## Proposed changes

This pr implements website support for profile pictures.

Including:
- Displaying and subscribing to changes
- Uploading profile pictures
- Removing profile pictures

## Types of changes

## Checklist

- [ ] Lint and unit tests pass locally with my changes

## Further comments


platform/api/src/api/v1/gql/models/user.rs
```diff
@@ -21,6 +21,7 @@ pub struct User<G: ApiGlobal> {
 	pub display_color: DisplayColor,
 	pub username: String,
 	pub channel: Channel<G>,
+	pub profile_picture_pending: bool,
```
TroyKomodo, line 24: Why not return the pending id here?

platform/image_processor/src/processor/job/mod.rs
```diff
@@ -129,8 +129,9 @@ impl<'a, G: ImageProcessorGlobal> Job<'a, G> {
 			}
 		}
 		// job completion
+		tracing::info!("publishing job completion event to {}", self.job.task.callback_subject);
 		self.global
-			.nats()
+			.jetstream()
```
TroyKomodo, line 134: This should be nats?

platform/website/src/lib/utils.ts
```diff
@@ -93,6 +93,19 @@ function formatSmallNumber(number: number): string {
...
+	return a.reduce((res: { [key: string]: T[] }, x) => {
+		const key = f(x);
+		if (!res[key]) {
+			res[key] = [];
+		}
+		res[key].push(x);
+		return res;
+	}, {});
+}
+
```
TroyKomodo, line 108: Do you want to add it to the object prototype or not?

TroyKomodo (approved):
looks good

platform/image_processor/src/processor/job/mod.rs
lennartkloock: I don't think so. Platform api uses jetstream to handle the callback. See [image_upload_callback.rs file](https://github.com/ScuffleTV/scuffle/blob/lennart/platform-profile-pictures/platform/api/src/image_upload_callback.rs#L24)

platform/website/src/lib/utils.ts
lennartkloock: I prefer not to since it can get confusing, especially because groupBy isn't the name in all browsers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy#browser_compatibility

IS2511:
The slash in parenthesis of the commit message is for sure not to format. Please choose one or remove the parenthesis entirely if the scope is global.

lennartkloock:
> The slash in parenthesis of the commit message is for sure not to format. Please choose one or remove the parenthesis entirely if the scope is global.

Because this is a mono repo the scope is not global most of the time. What scope would you suggest for `platform/website` or `platform/api`?

4w3Official (approved):
lgtm

IS2511:
@lennartkloock Hmm, I've googled around a bit and it looks like dashes are generally uhhh ok, probably.

Conventional Commits is strongly against anything but a single word:
> A scope MAY be provided after a type. A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., `fix(parser):`

Karma Style is indifferent:
> Example `<scope>` values:
> - `web-server`

I personally like the slash, it clearly conveys the intent - "part of" or a "subdirectory", instead of just "component name". A dash is a compromise I'm willing to accept (e.g. `platform-api`).

On a related note, I think the prefix of the commit message is getting too long with that subject name ij general. Maybe we should name the components somehow and use that shorter name? How about shortening `website` to `web`? Not much help though... Still, shorter is better in this case IMO.

IS2511:
I just noticed the `image_processor` commits. Not sure how I feel about that underscore... Probably negatively. If those will be squashed anyway then it doesn't matter much.

lennartkloock:
@IS2511 I'll just leave out `platform/` or `video/` in the future. I guess that's enough because apart from api all components have unique names anyway

TroyKomodo (approved):

PR #218, opened 2024-12-05
Add CI Workflows

TroyKomodo:
## Motivation

Add testing and code coverage to our CI runner so that we can work on code coverage and unit testing the libraries we write.

## Solution

Uses codecov and Just as our formatter.


TroyKomodo:
@codecov-ai-reviewer review

PR #228, opened 2025-01-03
add orphaned diff finder

TroyKomodo:
Adds a github action that finds differences in PRs that are not in the final PR diff.

For example if 

`Commit A` [**`adds random.bin`**](#)
`Commit B` [**`removes random.bin`**](#)
`Commit C` [**`changes x.py`**](#)

The final PR diff will only show the changes for `x.py` and the `random.bin` addition and removal will not be shown. This action finds and reports those and other line diffs so that we can understand if something has been secretly inserted into git history.

### DIFF
--- a/.github/workflows/orphaned.yaml
+++ b/.github/workflows/orphaned.yaml
@@ -1,0 +1,68 @@
+name: "PR Orphaned Diff Finder"
+
+on:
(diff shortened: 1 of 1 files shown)


TroyKomodo:
?brawl merge

PR #230, opened 2025-01-04
remove orphaned checker

TroyKomodo:
The checker needs more polish before adding, it often reports a lot of noise


TroyKomodo:
?brawl merge

PR #234, opened 2025-01-05
add future-ext

TroyKomodo:
I find the syntax for `tokio::time::timeout` to be quite ugly, since it breaks the typical chained method calls. With the FutureExt trait we can just call `.with_timeout(..)` instead which seems to be a cleaner way to do it.

### DIFF
--- a/crates/future-ext/src/lib.rs
+++ b/crates/future-ext/src/lib.rs
@@ -1,0 +1,34 @@
+/// The [`FutureExt`] trait is a trait that provides a more ergonomic way to
+/// extend futures with additional functionality. Similar to the [`IteratorExt`]
+/// trait, but for futures.
+pub trait FutureExt {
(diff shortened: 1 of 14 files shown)


crates/future-ext/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+/// The [`FutureExt`] trait is a trait that provides a more ergonomic way to
+/// extend futures with additional functionality. Similar to the [`IteratorExt`]
+/// trait, but for futures.
+pub trait FutureExt {
+    /// Attach a timeout to the future.
+    ///
+    /// The timeout is relative to the current time.
```
lennartkloock, line 7: This is confusing. The timeout is a duration.

lennartkloock (approved):

TroyKomodo:
?brawl merge

PR #235, opened 2025-01-05
fix doc coverage bug

TroyKomodo:
For some reason cargo llvm-cov report does not include doc coverage even if the doc coverage profraw files are there. It seems you also need to
 add the flag `--doctests` to the report command.

I opened an upstream issue to see if this is the desired behaviour or not. It seems a bit unintuitive.

https://github.com/taiki-e/cargo-llvm-cov/issues/412


lennartkloock (approved):

lennartkloock:
?brawl merge

PR #241, opened 2025-01-09
create scuffle-bytes-util

TroyKomodo:
Rename `bytesio` to `scuffle-bytes-util`. We also remove the wrapper types we had and optimize the `Cursor<Bytes>` usage. A lot of what we did with the wrapper types was essentially what `BytesMut` does and what `Vec<u8>` does, which is why most of the usage has been replaced with that.

These changes allow for the code to have fewer copies in the rtmp implementation (I will look to further clean up the tests and code there in a future PR). 

The old rtmp flow was such:

- copy bytes from wire into a temporary buffer
- copy that buffer into the chunk decoder (or handshake buffer) buffer
- parse chunks/handshake
- create a temporary buffer to write back result
- copy result from tmp buffer onto wire.

The new flow looks like:

- copy bytes from wire into a buffer
- parse chunks/handshake
- write result into a buffer
- copy result onto the wire

The newer example reuses existing buffers; so each connection will store additional memory (roughly 10kb/connection). As apposed to creating a buffer each time we want to read or write we just reuse an existing one. 

We also remove the additional copy to the handshake or chunk decode buffers by having them all share the same buffer (the read buffer).

CLOUD-23


TroyKomodo:
Build broken due to https://github.com/rust-lang/rust/issues/135235

crates/bytes-util/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "scuffle-bitio"
+name = "scuffle-bytes-util"
```
TroyKomodo, line 2: open to name suggestions? `scuffle-io-util`? not sure.

lennartkloock: I think `scuffle-bytes-util` is better than `scuffle-io-util`.

crates/bytes-util/src/bytes_cursor.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// This does not do a copy of the bytes, and is O(1) time.
+    ///
+    /// This is the same as `BytesCursor::extract_bytes(self.remaining())`.
+    fn extract_remaining(&mut self) -> Bytes;
+
+    /// Extracts a bytes from the cursor.
+    ///
+    /// This does not do a copy of the bytes, and is O(1) time.
+    /// Returns an error if the size is greater than the remaining bytes.
+    fn extract_bytes(&mut self, size: usize) -> io::Result<Bytes>;
```
TroyKomodo, line 21: Open to naming suggestions for these methods too. I dont really like them at the moment.

lennartkloock (approved):

TroyKomodo:
> You should change the shields.io links in the readme

which one?

lennartkloock (approved):
nvm I missed the change

crates/bytes-util/src/bytes_cursor.rs
```diff
@@ -0,0 +1,113 @@
+use std::io;
+
+use bytes::Bytes;
+
+/// A helper trait to implement zero copy reads on a `Cursor<Bytes>` type.
+pub trait BytesCursor {
+    /// Returns the remaining bytes in the cursor.
+    fn remaining(&self) -> usize;
+
+    /// Extracts the remaining bytes from the cursor returning.
```
lennartkloock, line 10: I feel like this sentence doesn't really make sense

crates/bytes-util/src/bytes_cursor.rs
```diff
@@ -0,0 +1,113 @@
...
+    ///
+    /// This does not do a copy of the bytes, and is O(1) time.
+    ///
+    /// This is the same as `BytesCursor::extract_bytes(self.remaining())`.
+    fn extract_remaining(&mut self) -> Bytes;
+
+    /// Extracts a bytes from the cursor.
```
lennartkloock, line 17: Remove "a"
```suggestion
    /// Extracts bytes from the cursor.
```

lennartkloock (changes requested):

crates/bytes-util/src/bytes_cursor.rs
lennartkloock: Why not `read_bytes`?

lennartkloock (approved):

lennartkloock:
?brawl merge

TroyKomodo:
?brawl retry

PR #249, opened 2025-01-11
improve docs for expgolomb crate

TroyKomodo:
Adds documentation and cleanup for the expgolomb crate. I was debating the naming because it frankly looks quite ugly, however I ended up just naming it `expgolomb` because I saw other libraries name it this. https://github.com/ramenhut/expgolomb (C++ Library)

There is another rust library that does this too https://docs.rs/exp-golomb/latest/exp_golomb/index.html

CLOUD-26


crates/expgolomb/src/lib.rs
```diff
@@ -0,0 +1,346 @@
+//! # scuffle-expgolomb
+//!
+//! [![crates.io](https://img.shields.io/crates/v/scuffle-expgolomb.svg)](https://crates.io/crates/scuffle-expgolomb) [![docs.rs](https://img.shields.io/docsrs/scuffle-expgolomb)](https://docs.rs/scuffle-expgolomb)
+//!
+//! ---
+//!
```
lennartkloock, line 6: This shouldn't be here

lennartkloock (changes requested):

lennartkloock (approved):

lennartkloock:
?brawl merge

PR #250, opened 2025-01-11
flv crate refactor

TroyKomodo:
Refactors the FLV Crate adding support for the new enhanced spec and improving the usage and API docs.

CLOUD-27


crates/flv/src/script.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub fn demux(reader: &mut io::Cursor<Bytes>) -> io::Result<Self> {
+        let buf = reader.extract_remaining();
+        let mut amf0_reader = Amf0Decoder::new(&buf);
+
+        let name = match amf0_reader.decode_with_type(Amf0Marker::String) {
+            Ok(Amf0Value::String(name)) => name,
+            _ => return Err(io::Error::new(io::ErrorKind::InvalidData, "Invalid script data name")),
+        };
+
+        let data = amf0_reader
```
lennartkloock, line 23: It is not clear to me from the spec whether this can only be an amf0 array value or if it can be any amf0 value (like implemented here)

crates/flv/src/file.rs
```diff
@@ -0,0 +1,41 @@
...
+            // backwards.
+            reader.read_u32::<BigEndian>()?;
+
+            // If there is no more data, we can stop reading.
+            if !reader.has_remaining() {
+                break;
+            }
+
+            // Demux the tag from the reader.
+            let tag = FlvTag::demux(reader)?;
```
lennartkloock, line 35: `FlvTag::demux` uses `extract_remaining` internally. This makes it impossible to read more than one tag. Am I missing something?

lennartkloock (changes requested):

crates/flv/src/script.rs
TroyKomodo: We just collect all remaining data into an array.

crates/flv/src/file.rs
TroyKomodo: Yea, so because we dont know how long the packet is; we just have to read EVERYTHING.

TroyKomodo:
```rs
        // We then extract the data from the reader. (advancing the cursor to the end of
        // the tag)
        let data = reader.extract_bytes(data_size as usize)?;

        // Finally we demux the data.
        let data = FlvTagData::demux(tag_type, &mut std::io::Cursor::new(data))?;
```

We do extract only what we need from the reader however

lennartkloock: Oh I missed that we create a new cursor there

crates/flv/src/video.rs
```diff
@@ -0,0 +1,266 @@
...
+
+impl VideoData {
+    pub fn demux(reader: &mut io::Cursor<Bytes>) -> io::Result<Self> {
+        let byte = reader.read_u8()?;
+        let enhanced = (byte & 0b1000_0000) != 0;
+        let frame_type_byte = (byte >> 4) & 0b0111;
+        let packet_type_byte = byte & 0b0000_1111;
+        let frame_type = FrameType::from(frame_type_byte);
+        let body = if frame_type == FrameType::VideoInfoOrCommandFrame {
+            let command_packet = CommandPacket::from(reader.read_u8()?);
```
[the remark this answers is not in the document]
lennartkloock: Talked in DMs, 10.1 spec seems to be wrong

lennartkloock (approved):

lennartkloock:
?brawl merge

PR #255, opened 2025-01-16
added test coverage for ffmpeg io

philipch07:
## Motivation

Added test coverage to ffmpeg io files.
## Solution

I added tests, and one function for the dictionary file for testing purposes (just to check if it's empty). Note that I was unable to figure out how to make tests for the `io/output.rs` file, namely the `write_xyz()` functions.

There was one strange case I ran into with the `flags()` function which returned a value of `2097152`. I did my best to look into it but I have no idea why it does that, so I left the test commented out in the file.
## Tickets

n/a

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -23,7 +23,7 @@ test *args:
     INSTA_FORCE_PASS=1 cargo +{{RUST_TOOLCHAIN}} llvm-cov nextest --include-build-script --no-report --all-features -- {{args}}
     # Coverage for doctests is currently broken in llvm-cov.
     # Once it fully works we can add the `--doctests` flag to the test and report command again.
-    cargo +{{RUST_TOOLCHAIN}} llvm-cov test --doc --no-report -- {{args}}
+    cargo +{{RUST_TOOLCHAIN}} llvm-cov test --doc --no-report --all-features -- {{args}}
 
     # Do not generate the coverage report on CI
     cargo insta review
--- a/Justfile
+++ b/Justfile
@@ -20,7 +20,7 @@ test *args:
     set -euo pipefail
 
     INSTA_FORCE_PASS=1 cargo +{{RUST_TOOLCHAIN}} llvm-cov clean --workspace
-    INSTA_FORCE_PASS=1 cargo +{{RUST_TOOLCHAIN}} llvm-cov nextest --include-build-script --no-report -- {{args}}
+    INSTA_FORCE_PASS=1 cargo +{{RUST_TOOLCHAIN}} llvm-cov nextest --include-build-script --no-report --all-features -- {{args}}
     # Coverage for doctests is currently broken in llvm-cov.
     # Once it fully works we can add the `--doctests` flag to the test and report command again.
     cargo +{{RUST_TOOLCHAIN}} llvm-cov test --doc --no-report -- {{args}}
--- a/crates/ffmpeg/src/io/channel.rs
+++ b/crates/ffmpeg/src/io/channel.rs
@@ -11,8 +11,6 @@ pub struct ChannelCompat<T: Send> {
     /// not sure. FFmpeg might require the IO to be synchronized, but I do not
     /// think it does.
     inner: Arc<Mutex<T>>,
-    total: usize,
-    pkt_idx: usize,
     buffer: BytesMut,
 }
 
@@ -20,8 +18,6 @@ impl<T: Send> ChannelCompat<T> {
     pub fn new(inner: T) -> Self {
         Self {
             inner: Arc::new(Mutex::new(inner)),
-            total: 0,
-            pkt_idx: 0,
             buffer: BytesMut::new(),
         }
     }
@@ -31,6 +27,8 @@ pub trait ChannelCompatRecv: Send {
     type Data: AsRef<[u8]>;
 
     fn channel_recv(&mut self) -> Option<Self::Data>;
+
+    fn try_channel_recv(&mut self) -> Option<Self::Data>;
 
     fn into_compat(self) -> ChannelCompat<Self>
     where
@@ -60,6 +58,10 @@ impl<D: AsRef<[u8]> + Send> ChannelCompatRecv for tokio::sync::mpsc::Receiver<D>
     fn channel_recv(&mut self) -> Option<Self::Data> {
         self.blocking_recv()
     }
+
+    fn try_channel_recv(&mut self) -> Option<Self::Data> {
+        self.try_recv().ok()
+    }
 }
 
 #[cfg(feature = "tokio-channel")]
@@ -77,6 +79,10 @@ impl<D: AsRef<[u8]> + Send> ChannelCompatRecv for tokio::sync::mpsc::UnboundedRe
 
     fn channel_recv(&mut self) -> Option<Self::Data> {
         self.blocking_recv()
+    }
+
+    fn try_channel_recv(&mut self) -> Option<Self::Data> {
+        self.try_recv().ok()
     }
 }
 
@@ -96,6 +102,10 @@ impl<D: AsRef<[u8]> + Clone + Send> ChannelCompatRecv for tokio::sync::broadcast
     fn channel_recv(&mut self) -> Option<Self::Data> {
         self.blocking_recv().ok()
     }
+
+    fn try_channel_recv(&mut self) -> Option<Self::Data> {
+        self.try_recv().ok()
+    }
 }
 
 #[cfg(feature = "tokio-channel")]
@@ -114,6 +124,10 @@ impl<D: AsRef<[u8]> + Send> ChannelCompatRecv for crossbeam_channel::Receiver<D>
     fn channel_recv(&mut self) -> Option<Self::Data> {
         self.recv().ok()
     }
+
+    fn try_channel_recv(&mut self) -> Option<Self::Data> {
+        self.try_recv().ok()
+    }
 }
 
 #[cfg(feature = "crossbeam-channel")]
@@ -131,6 +145,10 @@ impl<D: AsRef<[u8]> + Send> ChannelCompatRecv for std::sync::mpsc::Receiver<D> {
     fn channel_recv(&mut self) -> Option<Self::Data> {
         self.recv().ok()
     }
+
+    fn try_channel_recv(&mut self) -> Option<Self::Data> {
+        self.try_recv().ok()
+    }
 }
 
 impl<D: From<Vec<u8>> + Send> ChannelCompatSend for std::sync::mpsc::Sender<D> {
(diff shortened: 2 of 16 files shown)


crates/ffmpeg/src/io/input.rs
```diff
@@ -146,3 +146,125 @@ impl Input<()> {
...
+    }
+
+    #[test]
+    fn test_streams() {
+        let valid_file_path = "../../assets/avc_aac_large.mp4";
+        let input = Input::open(valid_file_path).expect("Failed to open valid file");
+
+        let streams = input.streams();
+        assert!(streams.len() > 0, "Expected at least one stream");
+    }
```
TroyKomodo, line 239: I think we should assert the actual underlying streams. 
We should check how many streams this has and that they are what we expect when decoding the avc_aac_large.mp4 file.

You can do this when you use `insta::assert_debug_snapshot` use the inline format
```
insta::assert_debug_snapshot(input.streams.collect::<Vec<_>>, @"")
```

So we can make sure the state does not change if it is correct now.

crates/ffmpeg/src/io/input.rs
```diff
@@ -146,3 +146,125 @@ impl Input<()> {
...
+        let mut packets = input.packets();
+
+        for _ in 0..5 {
+            match packets.next() {
+                Some(Ok(_)) => (),
+                Some(Err(e)) => panic!("Error encountered while reading packets: {:?}", e),
+                None => break,
+            }
+        }
+    }
```
TroyKomodo, line 254: Similarly with the packets, we should also assert their state does not change in future release.

crates/ffmpeg/src/io/input.rs
```diff
@@ -146,3 +146,125 @@ impl Input<()> {
...
+        match input.receive_packet() {
+            Ok(Some(packet)) => {
+                assert!(!packet.data().is_empty(), "Expected a non-empty packet");
+                assert!(packet.stream_index() >= 0, "Expected a valid stream index");
+            }
+            Ok(None) => panic!("Expected a packet but received None"),
+            Err(e) => panic!("Error encountered while receiving packet: {:?}", e),
+        }
+    }
+}
```
TroyKomodo, line 270: Also here.

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+        let mut opaque = Box::new(data);
+        let mut buf = [0u8; 10];
+
+        unsafe {
+            let result =
+                read_packet::<Cursor<Vec<u8>>>(&mut *opaque as *mut _ as *mut c_void, buf.as_mut_ptr(), buf.len() as i32);
+
+            assert_eq!(result, AVERROR_EOF);
+        }
+    }
```
TroyKomodo, line 288: This is unsafe & unsound I think.

I think all you need to do here is take the 
```
(&raw mut data) as *mut libc::c_void
```

You do not need to sue a box and deref the box.

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+
+        unsafe {
+            let result =
+                write_packet::<Cursor<Vec<u8>>>(&mut *opaque as *mut _ as *mut c_void, buf.as_ptr(), buf.len() as i32);
+            assert_eq!(result, buf.len() as i32);
+
+            let written_data = opaque.get_ref();
+            assert_eq!(&written_data[..buf.len()], &buf);
+        }
+    }
```
TroyKomodo, line 304: Same with these functions like above

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+
+    #[test]
+    fn test_seek_force() {
+        let mut cursor = Cursor::new(vec![0u8; 100]);
+        let opaque = &mut cursor as *mut _ as *mut c_void;
+        assert_eq!(cursor.position(), 0);
```
TroyKomodo, line 310: We should also assert the resulting cursor position after the seek.

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+        let result = unsafe { seek::<Cursor<Vec<u8>>>(opaque, offset, whence) };
+
+        assert_eq!(result, { offset });
+        whence &= !AVSEEK_FORCE;
+        assert_eq!(whence, SEEK_CUR);
+    }
+
+    #[test]
+    fn test_seek_seek_end() {
+        let mut cursor = Cursor::new(vec![0u8; 100]);
```
TroyKomodo, line 322: similarly with these seeks we should assert the cursor position after we do a seek.

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+        assert_eq!(result, 90);
+    }
+
+    #[test]
+    fn test_seek_invalid_whence() {
+        let mut cursor = Cursor::new(vec![0u8; 100]);
+        let opaque = &mut cursor as *mut _ as *mut libc::c_void;
+        let result = unsafe { seek::<Cursor<Vec<u8>>>(opaque, 0, 999) };
+
+        assert_eq!(result, -1);
```
TroyKomodo, line 337: We should check that the cursor did not change position

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+    }
+
+    #[test]
+    fn test_avformat_alloc_output_context2_error() {
+        unsafe extern "C" fn dummy_write_fn(_opaque: *mut libc::c_void, _buf: *const u8, _buf_size: i32) -> i32 {
+            0 // simulate success
+        }
```
TroyKomodo, line 344: We should assert the contents of the buf & buf_size fields. 

The trick here is you can use static variables to store state between runs. Perhaps a static AtomicBool so that u can guarantee that the function is called at most once?

Or if you expect the function to be called multiple times then n times.

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+            }
+        }
+    }
+
+    #[test]
+    fn test_open_output_valid_path() {
+        let test_path = "test_output_file.mp4";
+        if Path::new(test_path).exists() {
+            fs::remove_file(test_path).unwrap();
+        }
```
TroyKomodo, line 372: I feel like we should use a lib like `tmp_file` or something to automatically create a new file and then delete it on cleanup.

crates/ffmpeg/src/io/internal.rs
```diff
@@ -259,3 +259,145 @@ impl Inner<()> {
...
+        let result = Inner::open_output(test_path);
+        if let Err(error) = &result {
+            eprintln!("Function returned an error: {:?}", error);
+        }
+
+        assert!(
+            matches!(result, Err(FfmpegError::Code(_))),
+            "Expected FfmpegError::Code but received a different error."
+        );
+    }
```
TroyKomodo, line 402: Can you explain how this test works. I do not like that we write to a /root/ path here.

crates/ffmpeg/src/io/output.rs
```diff
@@ -281,3 +281,206 @@ impl Output<()> {
...
+    fn test_output_options_get_format_ffi() {
+        let options = OutputOptions::default().format_name("mp4");
+        let format_ffi = options.get_format_ffi();
+        assert!(format_ffi.is_ok());
+        assert!(!format_ffi.unwrap().is_null());
+    }
+
+    #[test]
+    fn test_output_options_get_format_ffi_null() {
+        let dummy_format_ffi: *const AVOutputFormat = 0x1234 as *const AVOutputFormat;
```
TroyKomodo, line 340: Can we not get a real format from somewhere else instead of casting an integer to a pointer?

crates/ffmpeg/src/io/output.rs
```diff
@@ -281,3 +281,206 @@ impl Output<()> {
...
+        let data = Cursor::new(Vec::new());
+        let options = OutputOptions::default().format_name("mp4");
+        let mut output = Output::new(data, options).expect("Failed to create Output");
+        // create a new output to prevent double mut borrow
+        let data = Cursor::new(Vec::new());
+        let options = OutputOptions::default().format_name("mp4");
+        let mut output_two = Output::new(data, options).expect("Failed to create Output");
+
+        let dummy_codec: *const AVCodec = 0x1234 as *const AVCodec;
+        let source_stream = output.add_stream(Some(dummy_codec)).expect("Failed to add source stream");
```
TroyKomodo, line 442: What does the source stream have pre-configured? I see you assert against time_base & start_time & disposition but like what values do they have when you by default create a stream. Can we modify them so that the copy actually checks if we copy the values and not just default init?

crates/ffmpeg/src/io/output.rs
```diff
@@ -281,3 +281,206 @@ impl Output<()> {
...
+    // for some reason the default flag value is 2097152, but I have no idea why.
+    // #[test]
+    // fn test_output_flags() {
+    //     let data = Cursor::new(Vec::new());
+    //     let options = OutputOptions::default().format_name("mp4");
+    //     let output = Output::new(data, options).expect("Failed to create
+    // Output");     let flags = output.flags();
+
+    //     assert_eq!(flags, 0, "Expected default flags to be 0");
+    // }
```
TroyKomodo, line 475: can you explain what you mean here?

crates/ffmpeg/src/io/output.rs
```diff
@@ -281,3 +281,206 @@ impl Output<()> {
...
+
+    #[test]
+    fn test_output_open() {
+        let temp_path = Path::new("test_output.mp4");
+        let _temp_file = File::create(temp_path).expect("Failed to create temporary file");
+        let output = Output::open(temp_path.to_str().unwrap());
+
+        assert!(output.is_ok(), "Expected Output::open to succeed");
+        std::fs::remove_file(temp_path).expect("Failed to remove temporary file");
+    }
```
TroyKomodo, line 485: We should use a tmp file again here.

TroyKomodo (commented):
Thanks this is a very good start to the testing for the ffmpeg crate!!!

crates/ffmpeg/src/io/internal.rs
philipch07: The test was aimed to hit the returned error in `Inner::open_output()`:
```rust
        // Safety: avio_open is safe to call
        let ec = unsafe { avio_open(&mut this.context.as_deref_mut_except().pb, path.as_ptr(), AVIO_FLAG_WRITE) };

        if ec != 0 {
            return Err(FfmpegError::Code(ec.into()));
        }
```
I've changed the test to instead use a tempfile instead of trying to access a restricted file in /root/.

Justfile
```diff
@@ -20,7 +20,7 @@ test *args:
     set -euo pipefail
 
     INSTA_FORCE_PASS=1 cargo +{{RUST_TOOLCHAIN}} llvm-cov clean --workspace
-    INSTA_FORCE_PASS=1 cargo +{{RUST_TOOLCHAIN}} llvm-cov nextest --include-build-script --no-report -- {{args}}
+    INSTA_FORCE_PASS=1 cargo +{{RUST_TOOLCHAIN}} llvm-cov nextest --include-build-script --no-report --all-features -- {{args}}
     # Coverage for doctests is currently broken in llvm-cov.
     # Once it fully works we can add the `--doctests` flag to the test and report command again.
     cargo +{{RUST_TOOLCHAIN}} llvm-cov test --doc --no-report -- {{args}}
```
TroyKomodo, line 26: Can we add the `--all-features` to the doc test too?

crates/ffmpeg/src/io/channel.rs
```diff
@@ -181,3 +181,255 @@ impl<T: ChannelCompatSend> std::io::Write for ChannelCompat<T> {
...
+        let compat = sender.into_compat();
+
+        assert!(compat.inner.lock().is_ok(), "Inner Mutex should be initialized and lockable");
+        assert_eq!(compat.total, 0, "Total should be initialized to 0");
+        assert_eq!(compat.pkt_idx, 0, "pkt_idx should be initialized to 0");
+        assert!(compat.buffer.is_empty(), "Buffer should be empty at initialization");
+    }
+
+    #[test]
+    fn test_mpsc_channel_send_recv() {
```
TroyKomodo, line 261: This test needs a feature flag so that it does not compile if tokio is not enabled.

crates/ffmpeg/src/io/channel.rs
```diff
@@ -181,3 +181,255 @@ impl<T: ChannelCompatSend> std::io::Write for ChannelCompat<T> {
...
+        // drop the sender to close the channel so .channel_recv() returns `None`
+        drop(tx_compat);
+
+        // channel is closed; second call to channel_recv should return None
+        let received_none = rx_compat.inner.lock().unwrap().channel_recv();
+        assert!(received_none.is_none());
+    }
+
+    #[test]
+    fn test_read_none_case() {
```
TroyKomodo, line 281: Similar to the above.

crates/ffmpeg/src/io/channel.rs
```diff
@@ -181,3 +181,255 @@ impl<T: ChannelCompatSend> std::io::Write for ChannelCompat<T> {
         Ok(())
     }
 }
+
+#[cfg(test)]
+#[cfg_attr(all(test, coverage_nightly), coverage(off))]
+mod tests {
```
TroyKomodo, line 187: I think for the channel tests, we should test generic features without tokio / crossbeam (only using std) and then specifically test tokio & crossbeam implementations with feature flags.

TroyKomodo (changes requested):

crates/ffmpeg/src/io/channel.rs
philipch07: I grouped the tests accordingly; the first few tests are those that test generic features, then there's a module for tests that require the `tokio-channel` feature flag, a module for tests that require the `crossbeam-channel` feature flags, and a final module for test that require both the `tokio-channel` and `crossbeam-channel` feature flags. Let me know what you think!

TroyKomodo:
really good work thanks @philipch07

TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #259, opened 2025-01-20
docs: clean up

lennartkloock:
- Removes headlines from crate root docs

Otherwise it would look like this when rendered:

headline shows twice

CLOUD-35


TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #262, opened 2025-01-21
pprof: add test and docs

lennartkloock:
- Adds a simple test
- Improves docs

CLOUD-11

### DIFF
--- a/crates/pprof/src/lib.rs
+++ b/crates/pprof/src/lib.rs
@@ -30,7 +30,8 @@ pub enum PprofError {
 #[cfg(test)]
 #[cfg_attr(all(coverage_nightly, test), coverage(off))]
 mod tests {
-    use std::{io::Read, time::SystemTime};
+    use std::io::Read;
+    use std::time::SystemTime;
 
     use flate2::read::GzDecoder;
     use pprof::protos::Message;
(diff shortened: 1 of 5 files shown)


TroyKomodo:
We should add a crate doc example

crates/pprof/src/lib.rs
```diff
@@ -1,5 +1,38 @@
...
+//! ```rust,no_run
+//! // Create a new CPU profiler with a sampling frequency of 1000 Hz and an empty ignore list.
+//! let cpu = scuffle_pprof::Cpu::new::<String>(1000, &[]);
+//!
+//! // Capture a pprof profile for 10 seconds.
+//! // This call is blocking. It is recommended to run it in a separate thread.
+//! let capture = cpu.capture(std::time::Duration::from_secs(10)).unwrap();
+//!
+//! // Write the profile to a file.
+//! std::fs::write("capture.pprof", capture).unwrap();
```
TroyKomodo, line 14: Perhaps also show some `go pprof` tool commands for generating a flame graph of the capture

TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #263, opened 2025-01-21
fix: rtmp test workaround

lennartkloock:
- Introduces a workaround for the failing rtmp test


TroyKomodo:
what if we just add a delay?

lennartkloock:
> what if we just add a delay?

Why do we even check if ffmpeg is done?

TroyKomodo:
Because ffmpeg should be done at that point i think

TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #265, opened 2025-01-24
http: rework, tests and docs

lennartkloock:
- Complete rewrite for `scuffle-http`
  - Missing webtransport support (should be done in a different pr)
- Make `scuffle-bootstrap-telemetry` use the new `scuffle-http` version
- Adds tests
- Adds docs

CLOUD-9 and CLOUD-10


crates/http/src/server/builder.rs
```diff
@@ -0,0 +1,351 @@
...
+                rustlsconfig.alpn_protocols.push(b"h2c".to_vec());
+            }
+
+            if self.enable_http3 {
+                rustlsconfig.alpn_protocols.push(b"h3".to_vec());
+            }
+        } else if self.enable_http3 {
+            // HTTP/3 doesn't work without TLS
+            return Err(ServerBuilderError::MissingRustlsConfig);
+        }
```
TroyKomodo, line 251: This shouldnt be an error, we should require that TLS be provided before you can enable http3. This should be done via the type system (using a stateful builder pattern)

.cargo/config.toml
```diff
@@ -1,2 +1,6 @@
 [alias]
 xtask = "run --package xtask --target-dir target/xtask -- "
+
+[build]
+# Required by the reqwest dependency in scuffle-http
```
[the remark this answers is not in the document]
lennartkloock: There is [this](https://doc.rust-lang.org/cargo/reference/config.html#targetcfgrustflags) but I tried it with `target.test` and it didn't work

crates/http/src/server/builder.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    /// Set the rustls configuration for the server.
+    ///
+    /// This enables TLS support for the server.
+    ///
+    /// Required for HTTP/3 support.
+    pub fn with_rustls(mut self, config: rustls::ServerConfig) -> Self {
+        self.rustls_config = Some(config);
+        self
+    }
```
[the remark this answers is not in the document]
lennartkloock: The reason I didn't use an `Arc` here is that we have to clone the config for each backend anyway since they all have to set their own `max_early_data_size` value on it.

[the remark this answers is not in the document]
TroyKomodo: What do you mean we have to clone anyways? And i think we should allow early data and should just handle the case when there is data read early.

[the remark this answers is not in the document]
lennartkloock: To set the `max_early_data_size` value we have to clone the config because we don't want to change the original config. I can look into handling early data.

crates/http/Cargo.toml
```diff
@@ -10,128 +10,69 @@ license = "MIT OR Apache-2.0"
...
-default = [
-    "http1",
-    "http2",
-    "tracing",
-]
+# https://doc.rust-lang.org/cargo/reference/features.html#dependency-features
+[features]
+default = ["http1", "http2", "tower"]
+# The hyper/tracing feature requires a custom build flag (see .cargo/config.toml)
+tracing = ["dep:tracing", "hyper?/tracing", "quinn?/log", "h3-quinn?/tracing", "h3?/tracing", "tokio-rustls?/logging", "rustls?/logging"]
```
TroyKomodo, line 59: Get rid of hyper tracing I don't want people to need to include a cfg flag for this. If you want to add hyper tracing add a new feature that enables that.

crates/http/src/backend/hyper/insecure.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        loop {
+            let _res: Result<_, Error<S>> = async {
+                let (tcp_stream, addr) = listener.accept().await?;
+
+                #[allow(unused_variables)]
+                let http1_enabled = false;
+                #[cfg(feature = "http1")]
+                let http1_enabled = self.http1_enabled;
+                #[allow(unused_variables)]
```
TroyKomodo, line 42: Just do the cfg(not(...))

crates/http/src/backend/hyper/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+            let body = crate::body::IncomingBody::from(body);
+            let req = http::Request::from_parts(parts, body);
+            http_service.call(req).await
+        }
+    });
+
+    tokio::spawn(async move {
+        let mut builder = auto::Builder::new(TokioExecutor::new());
+
+        let _res = if http1 && http2 {
```
TroyKomodo, line 52: How come there aren't any feature flags on the http1/http2 calls below?

crates/http/src/backend/hyper/secure.rs
```diff
@@ -0,0 +1,66 @@
...
+        // https://github.com/hyperium/hyper/issues/3841
+        rustls_config.max_early_data_size = 0;
+
+        let listener = tokio::net::TcpListener::bind(self.bind).await?;
+        let tls_acceptor = tokio_rustls::TlsAcceptor::from(Arc::new(rustls_config));
+
+        loop {
+            let _res: Result<_, Error<S>> = async {
+                let (tcp_stream, addr) = listener.accept().await?;
+                let stream = tls_acceptor.accept(tcp_stream).await?;
```
TroyKomodo, line 43: This should happen off the main accept thread. Otherwise if a client is slow at the tls handshake it blocks all other connections.

crates/http/src/backend/mod.rs
```diff
@@ -1,96 +1,10 @@
-use crate::svc::ConnectionAcceptor;
+//! This module contains the underlying backends for the server.
+//!
+//! You probably don't want to use this module directly and should instead use the [`HttpServer`](crate::HttpServer) struct.
```
TroyKomodo, line 3: Can we make it private?

crates/http/src/body.rs
```diff
@@ -1,149 +1,84 @@
...
-            #[cfg(feature = "h3")]
-            IncomingBodyInner::Quic(body) => body.is_end_stream(),
-            IncomingBodyInner::Empty => true,
+        #[allow(unreachable_patterns)]
+        match self {
+            #[cfg(any(feature = "http1", feature = "http2"))]
+            IncomingBody::Hyper(body) => body.is_end_stream(),
+            #[cfg(feature = "http3")]
+            IncomingBody::Quic(body) => body.is_end_stream(),
+            _ => false,
```
TroyKomodo, line 56: We should just put cfg(not(any(...)))

crates/http/src/body.rs
```diff
@@ -1,149 +1,84 @@
...
-            #[cfg(feature = "hyper")]
-            IncomingBodyInner::Tcp(body) => body.size_hint(),
-            #[cfg(feature = "h3")]
-            IncomingBodyInner::Quic(body) => body.size_hint(),
-            IncomingBodyInner::Empty => http_body::SizeHint::with_exact(0),
-        }
-    }
-}
-
-pin_project_lite::pin_project! {
```
TroyKomodo, line 82: I think we should keep this somehow it's useful to track how many bytes were sent down the wire.

crates/http/src/server/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+
+    fn get_available_addr() -> std::io::Result<std::net::SocketAddr> {
+        let listener = std::net::TcpListener::bind("127.0.0.1:0")?;
+        listener.local_addr()
+    }
+
+    const RESPONSE_TEXT: &str = "Hello, world!";
+
+    #[test]
+    fn missing_bind() {
```
TroyKomodo, line 369: Can we not also do this with the state builder?

crates/http/src/server/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+use crate::service::HttpServiceFactory;
+
+/// An error that can occur when building an [`HttpServer`](crate::HttpServer).
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+#[allow(clippy::enum_variant_names)]
+pub enum ServerBuilderError {
+    #[error("missing bind address")]
+    MissingBind,
+    #[error("missing service")]
+    MissingServiceFactory,
```
TroyKomodo, line 14: Is this needed if we have the statebuilder?

crates/http/src/server/builder.rs
```diff
@@ -0,0 +1,418 @@
...
+    /// Set the [`Context`](scuffle_context::Context) for the server.
+    ///
+    /// The server will terminate when the context is canceled.
+    ///
+    /// This is optional and defaults to the global context.
+    ///
+    /// # See Also
+    ///
+    /// - [`scuffle_context`]
+    pub fn with_ctx(mut self, ctx: scuffle_context::Context) -> Self {
```
TroyKomodo, line 140: Do we pass this context into the http requests? Do they live by this context?

crates/http/src/server/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+            let enable_http3 = self.enable_http3;
+
+            match (start_tcp_backend, enable_http3) {
+                #[cfg(feature = "http3")]
+                (false, true) => {
+                    use scuffle_context::ContextFutExt;
+
+                    let backend = crate::backend::h3::Http3Backend { bind: self.bind };
+
+                    match backend.run(self.service_factory, _rustls_config).with_context(self.ctx).await {
```
TroyKomodo, line 88: Here we add with_context but we never send the context into the future and therefore any Tokio spawns will not be bound by the context.

crates/http/src/service/tower_factory.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::IncomingRequest;
+
+/// A [`HttpServiceFactory`] that wraps a [`tower::MakeService`](https://docs.rs/tower/latest/tower/trait.MakeService.html).
+/// The given [`tower::MakeService`](https://docs.rs/tower/latest/tower/trait.MakeService.html) will be called to create a new service for each new connection.
+///
+/// Create by calling [`tower_make_service_factory`] or [`custom_tower_make_service_factory`].
+#[derive(Clone, Debug)]
+pub struct TowerMakeServiceFactory<M, T> {
+    make_service: M,
+    target: T,
```
TroyKomodo, line 13: Can you elaborate on why this is needed?

crates/http/Cargo.toml
TroyKomodo: Such as tracing-unstable

lennartkloock: I think it is useful to have tracing on dependencies. We openly say that this is just a wrapper around hyper and h3 so I think it should be fine.
Would people have to enable the cfg flag themselves when compiling? I was assuming that it would use the config from this repo to compile it.

crates/http/src/backend/hyper/mod.rs
lennartkloock: The whole module is annotated with `#[cfg(any(feature = "http1", feature = "http2"))]`

crates/http/src/backend/mod.rs
lennartkloock: I thought it might still be useful to someone who needs direct access to the non-builder api

crates/http/src/service/tower_factory.rs
lennartkloock: Every tower make service expects a target to create a new service. (See [here](https://docs.rs/tower/0.5.2/tower/trait.MakeService.html#tymethod.make_service)) You can simply use `()` as a target if you don't need it. That can be done using our `tower_make_service_factory` function. If you need a custom target you can use our `custom_tower_make_service_factory` function which allows you to specify one. If the target should be the connecting IP you can use our `tower_make_service_with_addr_factory` function.

crates/http/src/server.rs
```diff
@@ -4,25 +4,30 @@ use std::net::SocketAddr;
...
     #[cfg_attr(docsrs, doc(cfg(feature = "http1")))]
     enable_http1: bool,
+    #[builder(default = true)]
     #[cfg(feature = "http2")]
     #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
     enable_http2: bool,
+    #[builder(default = false)]
     #[cfg(feature = "http3")]
     #[cfg_attr(docsrs, doc(cfg(feature = "http3")))]
     enable_http3: bool,
```
TroyKomodo, line 33: so we know that http3 cannot be enabled without tls enabled. Therefore we should do something like
```rs
pub struct HttpServer<F> {
    // ...
    #[builder(setters(vis = "", name = enable_http3_internal))]
    enable_http3: bool,
    // ...
}
```

```rs
impl<F, S: http_server_builder::State> HttpServerBuilder<F, S> {
    pub fn enable_http3(
        self,
    ) -> HttpServerBuilder<http_server_builder::SetEnableHttp3<F, S>>
    where
        S::RustlsConfig: http_server_builder::IsSet,
    {
        self.enable_http3_internal()
    }
```

crates/http/src/server.rs
```diff
@@ -31,13 +36,78 @@ pub struct HttpServer<S> {
     rustls_config: Option<rustls::ServerConfig>,
```
TroyKomodo, line 34: Also is there no feature flag on this?

crates/http/README.md
```diff
@@ -7,16 +7,66 @@
...
+
+```rust
+# use scuffle_future_ext::FutureExt;
+# tokio_test::block_on(async {
+# let run = async {
+let service = scuffle_http::service::fn_http_service(|req| async move {
+    scuffle_http::http::Response::builder()
+    .status(scuffle_http::http::StatusCode::OK)
+    .header(scuffle_http::http::header::CONTENT_TYPE, "text/plain")
+    .body("Hello, world!".to_string())
```
TroyKomodo, line 44: Indents are a bit weird here.

crates/http/examples/src/axum.rs
```diff
@@ -0,0 +1,95 @@
...
+        .unwrap();
+
+    Ok(server_config)
+}
+
+// Load public certificate from file.
+fn load_certs(filename: &str) -> io::Result<Vec<CertificateDer<'static>>> {
+    // Open certificate file.
+    let certfile = fs::File::open(filename)
+        .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("failed to open {}: {}", filename, e)))?;
```
TroyKomodo, line 78: Can we use a path relative to the project `Cargo.toml`

crates/http/src/backend/h3/body.rs
```diff
@@ -1,88 +1,40 @@
...
-
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum State {
     Data(Option<u64>),
     Trailers,
     Done,
 }
 
-pub(crate) struct QuicIncomingBodyInner<B: BidiStream<Bytes>> {
+pub struct QuicIncomingBody<B: h3::quic::BidiStream<Bytes>> {
```
TroyKomodo, line 15: Do we ever use the B generic other then the `B::RecvStream`?  Should we maybe consider removing this bind and passing a recv stream directly rather than the bidi?

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        <S::Service as HttpService>::Error: std::error::Error + Debug + Display + Send + Sync,
+        <S::Service as HttpService>::ResBody: Send,
+        <<S::Service as HttpService>::ResBody as http_body::Body>::Data: Send,
+        <<S::Service as HttpService>::ResBody as http_body::Body>::Error: std::error::Error + Send + Sync,
+    {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("starting server");
+
+        // not quite sure why this is necessary but it is
+        rustls_config.max_early_data_size = u32::MAX;
```
TroyKomodo, line 37: I think at some later time we should allow this to be an `Arc` and handle the case where this is not set.

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::error::Error;
+use crate::service::{HttpService, HttpServiceFactory};
+
+pub mod body;
+mod utils;
+
+#[derive(Debug, Clone)]
+pub struct Http3Backend {
+    pub ctx: scuffle_context::Context,
+    pub bind: SocketAddr,
```
TroyKomodo, line 19: Why are these public?

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+                                Ok(Some((req, stream))) => {
+                                    let (mut send, recv) = stream.split();
+
+                                    let size_hint = req
+                                        .headers()
+                                        .get(http::header::CONTENT_LENGTH)
+                                        .and_then(|len| len.to_str().ok().and_then(|x| x.parse().ok()));
+                                    let body: QuicIncomingBody<BidiStream<_>> = QuicIncomingBody::new(recv, size_hint);
+                                    let req =
+                                        http::Request::from_parts(req.into_parts().0, crate::body::IncomingBody::from(body));
```
TroyKomodo, line 68: Use `req.map(|_| crate::body::IncomingBody::from(body));`

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+                                            if let Err(err) = _res {
+                                                tracing::warn!("error: {}", err);
+                                            }
+                                        }
+                                        .with_context(ctx.clone()),
+                                    );
+                                }
+                                // indicating no more streams to be received
+                                Ok(None) => {
+                                    break;
```
TroyKomodo, line 101: Is this line never hit?

crates/http/src/backend/h3/utils.rs
```diff
@@ -0,0 +1,27 @@
...
+    body: impl http_body::Body,
+) -> Result<(), h3::Error> {
+    let mut body = std::pin::pin!(body);
+    while let Some(frame) = std::future::poll_fn(|cx| body.as_mut().poll_frame(cx)).await {
+        match frame {
+            Ok(frame) => match frame.into_data().map_err(|f| f.into_trailers()) {
+                Ok(mut data) => send.send_data(data.copy_to_bytes(data.remaining())).await?,
+                Err(Ok(trailers)) => {
+                    send.send_trailers(trailers).await?;
+                    return Ok(());
```
TroyKomodo, line 16: Perhaps we have a test with trailers to make sure this works.

crates/http/src/backend/h3/utils.rs
```diff
@@ -0,0 +1,27 @@
...
+                }
+                Err(Err(_)) => continue,
+            },
+            Err(_) => return Ok(()),
```
TroyKomodo, line 20: Why does this error not matter?

TroyKomodo: Should we `finish` here?

crates/http/src/backend/hyper/insecure.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Debug, Clone)]
+pub struct InsecureBackend {
+    pub ctx: scuffle_context::Context,
+    pub bind: SocketAddr,
+    #[cfg(feature = "http1")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http1")))]
+    pub http1_enabled: bool,
+    #[cfg(feature = "http2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
+    pub http2_enabled: bool,
```
TroyKomodo, line 18: Why are these public?

crates/http/src/backend/hyper/insecure.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl InsecureBackend {
+    pub async fn run<F>(self, service_factory: F) -> Result<(), Error<F>>
```
TroyKomodo, line 22: we should add an instrument that has the field as `bind`.

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+mod utils;
+
+#[derive(Debug, Clone)]
+pub struct Http3Backend {
+    pub ctx: scuffle_context::Context,
+    pub bind: SocketAddr,
+}
+
+impl Http3Backend {
+    pub async fn run<S>(self, service_factory: S, mut rustls_config: rustls::ServerConfig) -> Result<(), Error<S>>
```
TroyKomodo, line 23: we should add an instrument that has the field as `bind`.

crates/http/src/backend/hyper/insecure.rs
```diff
@@ -0,0 +1,88 @@
...
+        <<F::Service as HttpService>::ResBody as http_body::Body>::Data: Send,
+        <<F::Service as HttpService>::ResBody as http_body::Body>::Error: std::error::Error + Send + Sync,
+    {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("starting server");
+
+        let listener = tokio::net::TcpListener::bind(self.bind).await?;
+
+        loop {
+            let (tcp_stream, addr) = match listener.accept().await {
```
TroyKomodo, line 38: Before we had the ability to run multiple connection accepts from different threads, can we add that too? worker count?

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        // not quite sure why this is necessary but it is
+        rustls_config.max_early_data_size = u32::MAX;
+        let crypto = h3_quinn::quinn::crypto::rustls::QuicServerConfig::try_from(rustls_config)?;
+        let server_config = h3_quinn::quinn::ServerConfig::with_crypto(Arc::new(crypto));
+
+        let endpoint = h3_quinn::quinn::Endpoint::server(server_config, self.bind)?;
+
+        // handle incoming connections and requests
+        while let Some(new_conn) = endpoint.accept().await {
```
TroyKomodo, line 44: Before we had the ability to run multiple connection accepts from different threads, can we add that too? worker count?

crates/http/src/backend/hyper/secure.rs
```diff
@@ -0,0 +1,112 @@
...
+#[derive(Debug, Clone)]
+pub struct SecureBackend {
+    pub ctx: scuffle_context::Context,
+    pub bind: SocketAddr,
+    #[cfg(feature = "http1")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http1")))]
+    pub http1_enabled: bool,
+    #[cfg(feature = "http2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
+    pub http2_enabled: bool,
```
TroyKomodo, line 19: Why are these public?

crates/http/src/backend/hyper/secure.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        // #[cfg(not(feature = "http1"))]
+        // let http1_enabled = false;
+        // #[cfg(feature = "http1")]
+        // let http1_enabled = self.http1_enabled;
+
+        // #[cfg(not(feature = "http2"))]
+        // let http2_enabled = false;
+        // #[cfg(feature = "http2")]
+        // let http2_enabled = self.http2_enabled;
```
TroyKomodo, line 44: Why are these commented out?

crates/http/src/backend/hyper/secure.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        // reset it back to 0 because everything explodes if it's not
+        // https://github.com/hyperium/hyper/issues/3841
+        rustls_config.max_early_data_size = 0;
+
+        let listener = tokio::net::TcpListener::bind(self.bind).await?;
+        let tls_acceptor = tokio_rustls::TlsAcceptor::from(Arc::new(rustls_config));
+
+        loop {
+            let (tcp_stream, addr) = match listener.accept().await {
```
TroyKomodo, line 54: Some as before we should have worker counts

crates/http/src/backend/hyper/secure.rs
```diff
@@ -0,0 +1,112 @@
...
+                }
+                #[cfg(not(feature = "tracing"))]
+                Err(_) => continue,
+            };
+
+            let tls_acceptor = tls_acceptor.clone();
+            let mut service_factory = service_factory.clone();
+            tokio::spawn(
+                async move {
+                    let stream = match tls_acceptor.accept(tcp_stream).await {
```
TroyKomodo, line 69: At some later point we should allow the tls config to be loaded after we do the initial handshake. this way we can handle multiple tls certificates from a single bind

crates/http/src/body.rs
```diff
@@ -4,82 +4,94 @@ use std::task::{Context, Poll};
...
+    type Error = IncomingBodyError;
+
+    fn is_end_stream(&self) -> bool {
+        #[allow(unreachable_patterns)]
+        match self {
+            #[cfg(any(feature = "http1", feature = "http2"))]
+            IncomingBody::Hyper(body) => body.is_end_stream(),
+            #[cfg(feature = "http3")]
+            IncomingBody::Quic(body) => body.is_end_stream(),
+            #[cfg(not(any(feature = "http1", feature = "http2", feature = "http3")))]
```
TroyKomodo, line 60: Can we know why none of these lines hit?

crates/http/src/body.rs
```diff
@@ -4,82 +4,94 @@ use std::task::{Context, Poll};
...
+    fn poll_frame(
+        self: std::pin::Pin<&mut Self>,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Result<http_body::Frame<Self::Data>, Self::Error>>> {
+        #[allow(unreachable_patterns)]
+        match self.get_mut() {
+            #[cfg(any(feature = "http1", feature = "http2"))]
+            IncomingBody::Hyper(body) => std::pin::Pin::new(body).poll_frame(_cx).map_err(Into::into),
+            #[cfg(feature = "http3")]
+            IncomingBody::Quic(body) => std::pin::Pin::new(body).poll_frame(_cx).map_err(Into::into),
```
TroyKomodo, line 74: This is never called?

crates/http/src/body.rs
```diff
@@ -4,82 +4,94 @@ use std::task::{Context, Poll};
...
-            IncomingBodyInner::Quic(body) => body.size_hint(),
-            IncomingBodyInner::Empty => http_body::SizeHint::with_exact(0),
+        #[allow(unreachable_patterns)]
+        match self {
+            #[cfg(any(feature = "http1", feature = "http2"))]
+            IncomingBody::Hyper(body) => body.size_hint(),
+            #[cfg(feature = "http3")]
+            IncomingBody::Quic(body) => body.size_hint(),
+            #[cfg(not(any(feature = "http1", feature = "http2", feature = "http3")))]
+            _ => http_body::SizeHint::default(),
```
TroyKomodo, line 88: This is also never called? Do we know why?

crates/http/src/body.rs
```diff
@@ -121,12 +144,12 @@ where
                 if let Some(Ok(frame)) = &frame {
                     if let Some(data) = frame.data_ref() {
                         if let Err(err) = this.tracker.on_data(data.remaining()) {
-                            return Poll::Ready(Some(Err(err.into())));
+                            return Poll::Ready(Some(Err(TrackedBodyError::Tracker(err))));
```
TroyKomodo, line 147: Can we add a test for this too?

crates/http/src/server.rs
```diff
@@ -0,0 +1,282 @@
...
+                    let hyper = std::pin::pin!(hyper);
+
+                    let mut http3 = crate::backend::h3::Http3Backend {
+                        ctx: self.ctx.clone(),
+                        bind: self.bind,
+                    }
+                    .run(self.service_factory, _rustls_config);
+                    let http3 = std::pin::pin!(http3);
+
+                    let res = futures::future::select(hyper, http3).with_context(self.ctx).await;
```
TroyKomodo, line 244: we shouldnt use `with_context` here because this will kill the select without a graceful shutdown

crates/http/Cargo.toml
```diff
@@ -10,128 +10,71 @@ license = "MIT OR Apache-2.0"
...
-    "http2",
-    "tracing",
-]
+# https://doc.rust-lang.org/cargo/reference/features.html#dependency-features
+[features]
+default = ["http1", "http2", "tower"]
+tracing = ["dep:tracing", "quinn?/log", "h3-quinn?/tracing", "h3?/tracing", "tokio-rustls?/logging", "rustls?/logging"]
+http1 = ["dep:hyper", "dep:hyper-util", "hyper/http1", "hyper-util/http1"]
+http2 = ["dep:hyper", "dep:hyper-util", "hyper/http2", "hyper-util/http2"]
+http3 = ["dep:quinn", "dep:h3-quinn", "dep:h3", "tls-rustls"]
```
TroyKomodo, line 64: I think this shouldn't add `tls-rustls` I think if someone use `http3` they should add both features manually. We should also add a `compile_error!` if `http3` is enabled without `tls-rustls` that way in the future we can have a alternative tls.

crates/http/src/backend/h3/body.rs
```diff
@@ -1,88 +1,40 @@
...
     stream: RequestStream<B::RecvStream, Bytes>,
     state: State,
 }
 
-impl<B: BidiStream<Bytes>> QuicIncomingBodyInner<B> {
-    #[cfg(feature = "quic-quinn")]
-    pub(crate) fn new(stream: RequestStream<B::RecvStream, Bytes>, size_hint: Option<u64>) -> Self {
+impl<B: h3::quic::BidiStream<Bytes>> QuicIncomingBody<B> {
+    pub fn new(stream: RequestStream<B::RecvStream, Bytes>, size_hint: Option<u64>) -> Self {
```
TroyKomodo, line 21: We should add a doc comment here.

crates/http/src/backend/h3/mod.rs
TroyKomodo: Also a doc comment

crates/http/src/backend/h3/utils.rs
```diff
@@ -0,0 +1,27 @@
+use bytes::{Buf, Bytes};
+use h3::quic::SendStream;
+use h3::server::RequestStream;
+
+pub async fn copy_response_body(
```
TroyKomodo, line 5: Add a doc comment

crates/http/src/backend/hyper/insecure.rs
TroyKomodo: doc comment

crates/http/src/backend/hyper/secure.rs
```diff
@@ -0,0 +1,112 @@
...
+    #[cfg(feature = "http1")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http1")))]
+    pub http1_enabled: bool,
+    #[cfg(feature = "http2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
+    pub http2_enabled: bool,
+}
+
+impl SecureBackend {
+    pub async fn run<F>(self, service_factory: F, mut rustls_config: rustls::ServerConfig) -> Result<(), Error<F>>
```
TroyKomodo, line 23: doc comment

crates/http/README.md
lennartkloock: Are they? What do you mean?

crates/http/src/backend/h3/mod.rs
lennartkloock: This can be used as a low-level API instead of the server builder pattern

crates/http/src/backend/hyper/insecure.rs
lennartkloock: This can be used as a low-level API instead of the server builder pattern

crates/http/src/backend/hyper/secure.rs
lennartkloock: This can be used as a low-level API instead of the server builder pattern

crates/http/src/backend/h3/mod.rs
lennartkloock: It's never hit but I'm not sure why

crates/http/src/backend/h3/body.rs
```diff
@@ -12,21 +12,21 @@ enum State {
     Done,
 }
 
-pub struct QuicIncomingBody<B: h3::quic::BidiStream<Bytes>> {
-    stream: RequestStream<B::RecvStream, Bytes>,
+pub struct QuicIncomingBody<S: h3::quic::RecvStream> {
```
TroyKomodo, line 15: We can get rid of the generic here since its not needed for type inference.

crates/http/src/backend/h3/body.rs
```diff
@@ -12,21 +12,21 @@ enum State {
...
+    stream: RequestStream<S, Bytes>,
     state: State,
 }
 
-impl<B: h3::quic::BidiStream<Bytes>> QuicIncomingBody<B> {
-    pub fn new(stream: RequestStream<B::RecvStream, Bytes>, size_hint: Option<u64>) -> Self {
+impl<S: h3::quic::RecvStream> QuicIncomingBody<S> {
```
TroyKomodo, line 20: Same with here.

crates/http/src/backend/h3/body.rs
```diff
@@ -12,21 +12,21 @@ enum State {
...
+    pub fn new(stream: RequestStream<S, Bytes>, size_hint: Option<u64>) -> Self {
         Self {
             stream,
             state: State::Data(size_hint),
         }
     }
 }
 
-impl<B: h3::quic::BidiStream<Bytes>> http_body::Body for QuicIncomingBody<B> {
+impl<S: h3::quic::RecvStream> http_body::Body for QuicIncomingBody<S> {
```
TroyKomodo, line 29: Keep this one since we need it for the function calls

crates/http/src/error.rs
```diff
@@ -1,14 +1,15 @@
...
-pub enum Error<S>
+pub enum Error<F>
 where
-    S: HttpServiceFactory,
-    S::Error: Debug + Display,
-    <S::Service as HttpService>::Error: Debug + Display,
+    F: HttpServiceFactory,
+    F::Error: Debug,
+    <F::Service as HttpService>::Error: Debug,
+    <<F::Service as HttpService>::ResBody as http_body::Body>::Error: Debug,
```
TroyKomodo, line 12: Is there any harm in requiring that this error be a `std::error::Error`? If not we should update it everywhere.

crates/http/src/backend/h3/utils.rs
```diff
@@ -30,6 +30,7 @@ where
             Ok(mut data) => send.send_data(data.copy_to_bytes(data.remaining())).await?,
             Err(Ok(trailers)) => {
                 send.send_trailers(trailers).await?;
+                send.finish().await?;
```
TroyKomodo, line 33: this is wrong. Send trailers does finish. You should not call both.

crates/http/src/backend/h3/mod.rs
```diff
@@ -31,7 +31,7 @@ impl Http3Backend {
     pub async fn run<F>(self, service_factory: F, mut rustls_config: rustls::ServerConfig) -> Result<(), Error<F>>
     where
         F: HttpServiceFactory + Clone + Send + 'static,
-        F::Error: Display + Debug + Send,
+        F::Error: std::error::Error + Display + Debug + Send,
```
TroyKomodo, line 34: Error requires display so we can drop that everywhere

crates/http/src/body.rs
lennartkloock: Not sure why

crates/http/src/body.rs
lennartkloock: The tests always run with all features enabled so this line is never hit

crates/http/examples/README.md
```diff
@@ -2,11 +2,11 @@
 
 Examples of using the `scuffle-http` crate.
 
-- [tcp](./tcp) - An example of how to use the crate to make a simple HTTP server.
+- [axum](./src/axum.rs) - Example of creating a cpu profile.
```
TroyKomodo, line 5: Add the other example here too

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+        rustls_config.max_early_data_size = u32::MAX;
+        let crypto = h3_quinn::quinn::crypto::rustls::QuicServerConfig::try_from(rustls_config)?;
+        let server_config = h3_quinn::quinn::ServerConfig::with_crypto(Arc::new(crypto));
+
+        let endpoint = h3_quinn::quinn::Endpoint::server(server_config, self.bind)?;
+
+        let tasks = (0..self.worker_tasks).map(|n| {
+            let ctx = self.ctx.clone();
+            let service_factory = service_factory.clone();
+            let endpoint = endpoint.clone();
```
TroyKomodo, line 57: This uses an Mutex under the hood so this does not actually work as expected when using multiple workers. (ie, multi thread accept)

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+                                                .get(http::header::CONTENT_LENGTH)
+                                                .and_then(|len| len.to_str().ok().and_then(|x| x.parse().ok()));
+                                            let body = QuicIncomingBody::new(recv, size_hint);
+                                            let req = req.map(|_| crate::body::IncomingBody::from(body));
+
+                                            let ctx = ctx.clone();
+                                            let mut http_service = http_service.clone();
+                                            tokio::spawn(async move {
+                                                // The context must live as long as the request is being handled
+                                                let _ctx = ctx;
```
TroyKomodo, line 103: I would replace this in favour of a `drop(ctx)` at the very end of this function.

crates/http/src/backend/h3/utils.rs
```diff
@@ -0,0 +1,43 @@
...
+    while let Some(frame) = std::future::poll_fn(|cx| body.as_mut().poll_frame(cx)).await {
+        match frame
+            .map_err(crate::error::Error::ResBodyError)?
+            .into_data()
+            .map_err(|f| f.into_trailers())
+        {
+            Ok(mut data) => send.send_data(data.copy_to_bytes(data.remaining())).await?,
+            Err(Ok(trailers)) => {
+                send.send_trailers(trailers).await?;
+                return Ok(());
```
TroyKomodo, line 34: We should add ut's to cover trailers on http3

crates/http/src/backend/hyper/insecure.rs
```diff
@@ -0,0 +1,149 @@
...
+                loop {
+                    #[cfg(feature = "tracing")]
+                    tracing::trace!("waiting for connections");
+
+                    let (tcp_stream, addr) = match listener.accept().with_context(ctx.clone()).await {
+                        Some(Ok(conn)) => conn,
+                        #[cfg(feature = "tracing")]
+                        Some(Err(e)) => {
+                            tracing::warn!(err = %e, "failed to accept tcp connection");
+                            continue;
```
TroyKomodo, line 69: Some errors here are fatal.

crates/http/src/backend/hyper/insecure.rs
TroyKomodo, line 1: I think we should get rid of `insecure` and `secure` backends and just have a `HyperBackend` where we handle tls and non tls on the same backend. Making a builder take a tls config as part of the build.

crates/http/src/backend/hyper/insecure.rs
```diff
@@ -0,0 +1,149 @@
...
+use crate::error::Error;
+use crate::service::{HttpService, HttpServiceFactory};
+
+/// A backend that handles incoming HTTP (no TLS) connections.
+///
+/// This is used internally by the [`HttpServer`](crate::server::HttpServer) but can be used directly if preferred.
+///
+/// Call [`run`](InsecureBackend::run) to start the server.
+#[derive(Debug, Clone)]
+pub struct InsecureBackend {
```
TroyKomodo, line 17: We should make all fields private and use bon to build.

TroyKomodo, line 17: We should make all fields private and use bon to build.

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+///
+/// This is used internally by the [`HttpServer`](crate::server::HttpServer) but can be used directly if preferred.
+///
+/// Call [`run`](Http3Backend::run) to start the server.
+#[derive(Debug, Clone)]
+pub struct Http3Backend {
+    pub ctx: scuffle_context::Context,
+    pub worker_tasks: usize,
+    pub bind: SocketAddr,
+}
```
TroyKomodo, line 27: We should use bon builder and make all fields private

changes.d/pr-265.toml
```diff
@@ -0,0 +1,5 @@
+[[scuffle-http]]
```
TroyKomodo, line 1: Can we add change logs for all the crates you update. Such as context and h3-webtransport.

crates/context/src/lib.rs
```diff
@@ -130,6 +130,12 @@ impl Clone for Context {
     }
 }
 
+impl Default for Context {
+    fn default() -> Self {
+        Self::global()
+    }
+}
```
TroyKomodo, line 137: Is this change required?

crates/http/README.md
```diff
@@ -7,16 +7,67 @@
...
+- `http3-tls-rustls`: Enables both `http3` and `tls-rustls` features. Disabled by default.
+
+## Example
+
+The following example demonstrates how to create a simple HTTP server (without TLS) that responds with "Hello, world!" to all requests on port 3000.
+
+```rust
+# use scuffle_future_ext::FutureExt;
+# tokio_test::block_on(async {
+# let run = async {
```
TroyKomodo, line 40: in the readme we can remove this bit

crates/http/README.md
```diff
@@ -7,16 +7,67 @@
...
+let service = scuffle_http::service::fn_http_service(|req| async move {
+    scuffle_http::http::Response::builder()
```
TroyKomodo, line 42: perhaps we should re-export the response type at the crate root

crates/http/src/body.rs
TroyKomodo: Can we figure out why?

crates/context/src/lib.rs
TroyKomodo: I think we should remove this tbh. I dont think Context default should be implemented.

changes.d/pr-265.toml
```diff
@@ -0,0 +1,10 @@
...
+breaking = true
+category = "feat"
+description = "complete rewrite with improved API"
+authors = ["@lennartkloock"]
+
+[[scuffle-bootstrap-telemetry]]
+category = "fix"
```
TroyKomodo, line 8: More like chore? Since you are bumping a dep that isn't exposed I think.

crates/http/src/backend/hyper/handler.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::error::Error;
+use crate::service::{HttpService, HttpServiceFactory};
+
+/// Helper function used by hyper server to handle incoming connections.
+pub async fn handle_connection<F, S, I>(
+    ctx: scuffle_context::Context,
+    service: S,
+    io: I,
+    #[cfg(feature = "http1")] http1: bool,
+    #[cfg(feature = "http2")] http2: bool,
```
TroyKomodo, line 15: We should not change the function signature depending on if these are enabled or not I think. Instead we should just accept both and do nothing if the feature is off?

crates/http/src/backend/hyper/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+
+                        #[cfg(all(feature = "http1", feature = "http2"))]
+                        let _res = handler::handle_connection::<F, _, _>(
+                            ctx,
+                            http_service,
+                            stream,
+                            self.http1_enabled,
+                            self.http2_enabled,
+                        )
+                        .await;
```
TroyKomodo, line 171: That way this logic becomes much simpler.

crates/http/src/body.rs
```diff
@@ -93,40 +103,70 @@ impl<B, T> TrackedBody<B, T> {
...
 
+    /// Called when data is read from the body.
+    ///
+    /// The `size` parameter is the size of the data that is remaining to be read from the body.
     fn on_data(&self, size: usize) -> Result<(), Self::Error> {
         let _ = size;
         Ok(())
     }
-
-    fn on_close(&self) {}
```
TroyKomodo, line 104: Why did we get rid of this method?

crates/http/src/backend/h3/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+
+pub mod body;
+mod utils;
+
+/// A backend that handles incoming HTTP3 connections.
+///
+/// This is used internally by the [`HttpServer`](crate::server::HttpServer) but can be used directly if preferred.
+///
+/// Call [`run`](Http3Backend::run) to start the server.
+#[derive(bon::Builder, Debug, Clone)]
```
TroyKomodo, line 23: We should add comments to each field so that the builder displays the comments on docs.

crates/http/src/backend/hyper/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+mod stream;
+mod utils;
+
+/// A backend that handles incoming HTTP connections using a hyper backend.
+///
+/// This is used internally by the [`HttpServer`](crate::server::HttpServer) but can be used directly if preferred.
+///
+/// Call [`run`](HyperBackend::run) to start the server.
+#[derive(Debug, Clone, bon::Builder)]
+pub struct HyperBackend<F> {
```
TroyKomodo, line 21: Likewise with the h3 builder we should add comments to each field to display them in the docs

crates/http/src/server.rs
```diff
@@ -0,0 +1,283 @@
...
+
+/// The HTTP server.
+///
+/// This struct is the main entry point for creating and running an HTTP server.
+///
+/// Start creating a new server by calling [`HttpServer::builder`].
+#[derive(Debug, Clone, bon::Builder)]
+#[builder(state_mod(vis = "pub(crate)"))]
+#[allow(dead_code)]
+pub struct HttpServer<F> {
```
TroyKomodo, line 15: Likewise with the other 2 builders, we should document each field.

crates/http/src/server.rs
```diff
@@ -0,0 +1,283 @@
...
+
+#[cfg(feature = "http3")]
+#[cfg_attr(docsrs, doc(cfg(feature = "http3")))]
+impl<F, S> HttpServerBuilder<F, S>
+where
+    S: http_server_builder::State,
+    S::EnableHttp3: http_server_builder::IsUnset,
+    S::RustlsConfig: http_server_builder::IsSet,
+{
+    pub fn enable_http3(self, enable_http3: bool) -> HttpServerBuilder<F, http_server_builder::SetEnableHttp3<S>> {
```
TroyKomodo, line 47: We should add a document comment here

crates/http/src/body.rs
lennartkloock: It was never called anywhere

crates/http/src/backend/hyper/handler.rs
```diff
@@ -0,0 +1,76 @@
...
+        tracing::warn!("both http1 and http2 are disabled, closing connection");
+        return Ok(());
+    };
+
+    #[cfg(feature = "tracing")]
+    if let Some(Err(e)) = _res {
+        tracing::warn!(err = %e, "connection error");
+    }
+
+    Ok(())
```
TroyKomodo, line 75: why dont we bubble this error?

crates/http/src/backend/hyper/handler.rs
```diff
@@ -0,0 +1,76 @@
...
+        builder
+            .serve_connection_with_upgrades(io, hyper_proxy_service)
+            .with_context(ctx)
+            .await
+    } else if http1 {
+        builder
+            .http1_only()
+            .serve_connection_with_upgrades(io, hyper_proxy_service)
+            .with_context(ctx)
+            .await
```
TroyKomodo, line 57: this should be feature gated

TroyKomodo:
?brawl try

crates/http/src/backend/hyper/handler.rs
```diff
@@ -50,12 +50,20 @@ where
             .with_context(ctx)
             .await
     } else if http1 {
+        #[cfg(not(feature = "http1"))]
+        return Ok(());
```
TroyKomodo, line 54: this can just be `unreachable!`

TroyKomodo (approved):

TroyKomodo:
?brawl merge

TroyKomodo:
?brawl cancel

TroyKomodo:
?brawl merge

TroyKomodo:
?brawl try

TroyKomodo:
?brawl cancel

TroyKomodo:
?brawl merge

lennartkloock:
?brawl try

TroyKomodo:
?brawl merge

PR #273, opened 2025-02-09
fix: memory leaks in ffmpeg tests

TroyKomodo:
Fixes 2 memory leaks in unit tests.

found using:

`CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="valgrind --error-exitcode=1 --leak-check=full" cargo nextest run -p scuffle-ffmpeg`

Rust does not run destructors of static variables therefore we need to manually unset the log callbacks when we add them.


lennartkloock (approved):

TroyKomodo:
?brawl merge

PR #287, opened 2025-02-11
CI Improvements

TroyKomodo:
Moves the clippy powerset stuff into the clippy job (only runs on a brawl run) and then adds a new job that runs `cargo package`. Cargo Package is a tool ran before `cargo publish` (it runs automatically unless you do `--no-verify`). This is a sanity check to test if the crate compiles before uploading it to crates.io. The reason we do this in a power-set is because we want to make sure that all features of the crate compile successfully. This is a known issue on crates.io where sometimes crates break and or dont work depending on the features they are enabled with.

CLOUD-63

This PR also splits our ci into components, to reuse existing setup code related to `setup-rust` where we install the rust tool chain, use a cache and install custom rust tools. As well as `setup-ffmpeg` where we install and configure ffmpeg.

CLOUD-64

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,14 +43,14 @@ jobs:
         if: ${{ github.event_name == 'push' }}
         run: cargo hakari disable
 
-      - name: Make sure code is linted
+      - name: Run Clippy
         if: ${{ github.event_name == 'pull_request' }}
         run: |
           set -xeo pipefail
           export RUSTFLAGS="-Dwarnings"
           cargo +${{ env.RUST_TOOLCHAIN }} clippy -Z unstable-options --all-features --all-targets --no-deps
 
-      - name: Make sure code is linted
+      - name: Run Clippy on Powerset
         if: ${{ github.event_name == 'push' }}
         run: |
           set -xeo pipefail
@@ -268,7 +268,7 @@ jobs:
       - name: Disable Hakari
         run: cargo hakari disable
 
-      - name: Run Powerset
+      - name: Run Cargo Package on Powerset
         run: just powerset package -- --allow-dirty
 
   brawl-done:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,26 +28,16 @@ jobs:
         with:
           ref: ${{ env.SHA }}
 
-      - name: Setup system deps
-        run: |
-          set -xeo pipefail
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends nasm libx264-dev
+      - name: Setup FFmpeg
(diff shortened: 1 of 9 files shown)


TroyKomodo:
?brawl try

.github/workflows/ci.yaml
```diff
@@ -45,9 +45,22 @@ jobs:
...
+        if: ${{ github.event_name == 'push' }}
+        run: cargo hakari disable
+
       - name: Make sure code is linted
+        if: ${{ github.event_name == 'pull_request' }}
         run: cargo +${{ env.RUST_TOOLCHAIN }} clippy --all-features --all-targets
 
+      - name: Make sure code is linted
+        if: ${{ github.event_name == 'push' }}
+        run: just powerset clippy
```
philipch07, line 62: Maybe give more descriptive names for these

.github/workflows/ci.yaml
```diff
@@ -332,11 +344,11 @@ jobs:
 
       - name: Run Powerset
         run: |
-          just powerset clippy
+          just powerset package -- --allow-dirty
```
philipch07, line 348: Should we rename this from "Run Powerset"?

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

.github/workflows/ci.yaml
TroyKomodo: This is fine I think its in the package workflow

.github/workflows/ci.yaml
TroyKomodo: Yes.

crates/bootstrap/telemetry/src/lib.rs
```diff
@@ -426,13 +426,21 @@ async fn opentelemetry_flush<G: TelemetryConfig>(
 
 #[cfg(test)]
 #[cfg_attr(all(test, coverage_nightly), coverage(off))]
+#[cfg(all(
+    feature = "opentelemetry-metrics",
+    feature = "opentelemetry-traces",
+    feature = "opentelemetry-logs"
+))]
```
lennartkloock, line 433:
```suggestion
#[cfg(feature = "opentelemetry")]
```

lennartkloock (changes requested):

lennartkloock (approved):

crates/bootstrap/telemetry/src/lib.rs
lennartkloock: nvm

TroyKomodo:
?brawl merge

PR #313, opened 2025-02-18
switch to renovate

TroyKomodo:
try using renovate


philipch07:
What's the difference between the dependency bot and renovate bot?

TroyKomodo:
it seems renovate does what i want it to do when we update deps for libraries. Just update the lock file. dependabot seems to want to update the version listed which is incorrect.

philipch07 (approved):

TroyKomodo:
?brawl merge

PR #365, opened 2025-02-19
bump rand 0.8 -> 0.9

TroyKomodo:
bumps the rand crate from 0.8 to 0.9 fixing all breaking changes

### DIFF
--- a/crates/pprof/examples/src/cpu.rs
+++ b/crates/pprof/examples/src/cpu.rs
@@ -3,7 +3,7 @@ use std::io::Write;
 use rand::Rng;
 
 fn work() {
-    let mut rnd = rand::thread_rng();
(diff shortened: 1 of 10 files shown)


crates/pprof/examples/src/cpu.rs
```diff
@@ -3,7 +3,7 @@ use std::io::Write;
 use rand::Rng;
 
 fn work() {
-    let mut rnd = rand::thread_rng();
+    let mut rnd = rand::rng();
```
philipch07, line 6: Thoughts on renaming the var names from `rnd` to `rng` to be consistent with the other uses of `rng` in the other files?

philipch07 (approved):

PR #366, opened 2025-02-19
bump config from 0.14 -> 0.15

TroyKomodo:
bumps config from 0.14 to 0.15

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -574,18 +574,18 @@ dependencies = [
 
 [[package]]
 name = "config"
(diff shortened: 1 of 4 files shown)


philipch07 (approved):
Was this left out of the automatic updates, and if so, is there a way for us to include it so we don't have to manually update this in the future?

TroyKomodo:
This was left out because its a breaking change on our end. We have to manually update it since they bumped a major version.

TroyKomodo:
?brawl merge

PR #370, opened 2025-02-19
h264 add docs and tests

philipch07:
Adds docs and updates/improves tests for h264.

CLOUD-29


philipch07:
?brawl try (I want to see if valgrind runs without a problem since I had to fix something locally to get it to work)

crates/h264/src/lib.rs
```diff
@@ -1,8 +1,46 @@
...
+//!
+//! This project is licensed under the [MIT](./LICENSE.MIT) or [Apache-2.0](./LICENSE.Apache-2.0) license.
+//! You can choose between one of them if you use this work.
+//!
+//! `SPDX-License-Identifier: MIT OR Apache-2.0`
+#![cfg_attr(all(coverage_nightly, test), feature(coverage_attribute))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![deny(missing_docs)]
+#![deny(clippy::undocumented_unsafe_blocks)]
+#![deny(clippy::multiple_unsafe_ops_per_block)]
```
TroyKomodo, line 37: Since this crate does not use unsafe we can just do `#[deny(unsafe)]`

crates/h264/README.md
```diff
@@ -7,7 +7,25 @@
...
+
+This crate is designed to provide a simple and safe interface to encode and decode H.264 headers.
+
+## Why do we need this?
+
+This crate aims to provides a simple and safe interface for h264.
+
+## How is this different from other h264 crates?
+
+The other main h264 crate is TODO.
```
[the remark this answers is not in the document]
philipch07: This and the below comment were the things I was asking if you could help with yesterday if you have time

crates/h264/src/config.rs
```diff
@@ -7,30 +7,48 @@ use bytes::{Buf, Bytes};
...
     pub level_indication: u8,
+    /// The `length_size_minus_one` is the u8 length of the NALUnitLength minus one.
     pub length_size_minus_one: u8,
+    /// The `sps` is a vec of SPS, each of which is a u64.
+    /// Refer to the SPS struct in the SPS docs for more info.
     pub sps: Vec<Bytes>,
+    /// The `pps` is a vec of PPS, each of which is a u64.
+    /// These contain syntax elements that can apply layer repesentation(s).
+    /// Note that they are supposed to be ordered by ascending PPS ID.
     pub pps: Vec<Bytes>,
```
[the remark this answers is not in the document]
philipch07: I remember that I thought the sizes were u64 from looking at our impl (double checking made me realize that I misread it as well). The exact type isn't mentioned in the docs, from what I can tell, but the variables are are referred in 5.3.2.1.3 of 11496-15:2022 which tells the reader to look at 14496-10 for `sequenceParameterSetNALUnit` and `pictureParameterSetNALUnit`, which I found the first definition at 3.1.141 and the second definition at 3.1.109.

crates/h264/src/sps.rs
```diff
@@ -6,40 +6,116 @@ use scuffle_bytes_util::BitReader;
 use scuffle_expgolomb::BitReaderExpGolombExt;
 
 #[derive(Debug, Clone, PartialEq)]
-/// Sequence parameter set
+/// The Sequence Parameter Set.
```
TroyKomodo, line 9: i think we should update these comments to reflect what they actually represent in the bit stream. This struct isnt a full decode of the sps only a partial with things I THINK are important. We can maybe implement the full decode of the sps in which case we should add a corresponding mux function too.

for example there is no such thing as the `frame_rate` in the sps standard but you can derive the frame_rate from some of the fields. 

so if we were to decode the entire sps we would do the frme rate derive thing as a method call instead of just always computing it.

crates/h264/src/sps.rs
```diff
@@ -6,40 +6,116 @@ use scuffle_bytes_util::BitReader;
...
     pub height: u64,
+    /// The framerate as a f64.
     pub frame_rate: f64,
+    /// An optional `ColorConfig`. Refer to the ColorConfig struct for more info.
     pub color_config: Option<ColorConfig>,
 }
 
 #[derive(Debug, Clone, PartialEq)]
-/// Color config for SPS
+/// The color config for SPS.
```
TroyKomodo, line 84: Likewise with this color config im not sure if this is the full struct, so we should update the field comments to explain where they are from or to parse the entire thing out. Also some of the u8s may not be full u8s and rather only be a few bits in which case we should say the amount of bits each field is.

crates/h264/src/sps.rs
```diff
@@ -205,15 +281,20 @@ impl Sps {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-/// Sequence parameter set extension.
+/// The Sequence Parameter Set extension.
```
TroyKomodo, line 284: The following fields are variable length integers we should specify that somehow.

crates/ffmpeg/src/io/snapshots/scuffle_ffmpeg__io__output__tests__test_output_write_mp4_fragmented_header.snap
```diff
@@ -1,7 +1,6 @@
 ---
 source: crates/ffmpeg/src/io/output.rs
 expression: get_boxes!(output)
-snapshot_kind: text
```
TroyKomodo, line 4: Any idea why this changed?

crates/h264/src/sps.rs
TroyKomodo: When I wrote this all I wanted to do was parse the frame rate because that's what I needed. So what my original comment was saying is that if we want to make this a full sps (we parse out the entire thing) then we should save all those other bits to the struct. Otherwise we shouldn't and just explain what our watered down version represents when it comes to the bits of the sps

crates/ffmpeg/src/io/snapshots/scuffle_ffmpeg__io__output__tests__test_output_write_mp4_fragmented_header.snap
philipch07: All I did was change the `chroma_format` field to be `chroma_format_idc` to match the spec and ran `just test` to update the snapshots. Maybe insta updated and they made it so it doesn't require that line anymore?

TroyKomodo: But these are from ffmpeg

crates/h264/src/sps.rs
philipch07: I see. Do you think users would find it useful to have an entirely parsed out SPS?

TroyKomodo: I think it would be ideal but might be a lot of effort. You are welcome to try it see if it's possible to represent the full sps in rust.

crates/ffmpeg/src/io/snapshots/scuffle_ffmpeg__io__output__tests__test_output_write_mp4_fragmented_header.snap
philipch07: Hm, I just checked on the main branch and the insta version for ffmpeg is 1.42, and the insta version for h264 is also 1.42. I'm not sure why it changed since I didn't manually change the snapshot.

crates/h264/src/sps.rs
philipch07: I'll see what I can do then! Would you mind checking the other rust h264 crates and see what they do?

crates/h264/src/config.rs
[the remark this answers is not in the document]
philipch07: I removed the size reference since they are both definitely not u64s.

crates/h264/README.md
```diff
@@ -7,7 +7,231 @@
...
+## How is this different from other h264 crates?
+
+There are currently a handful of other active crates:
+
+- Retina
+  - This crate is heavily focused towards being as lightweight as possible since it's meant for security cameras. The downside of this is that it lacks some SPS fields. It is also lacking some documentation for the specific SPS fields.
+- less-avc
+  - This crate is focused on being a simple pure Rust implementation of H.264 but with only a subset of its features. IT lacks 4:4:4 support and some documentation for the specific SPS fields.
+- openh264-rs
+  - This crate is a set of bindings around the C implementation of H.264 (openh264). As a result, the crate contains a lot of unsafe code since it's just a lightweight rust wrapper.
```
TroyKomodo, line 27: These 3 libraries are for either encoding / decoding the video bitstream. Currently our library is only for decoding the headers.

crates/h264/README.md
```diff
@@ -7,7 +7,231 @@
...
+
+## Notable features
+
+This crate is a completely safe implementation of H264 encoding and decoding, which means there is no unsafe code!
```
TroyKomodo, line 31: header encoding / decoding not the video stream.

crates/h264/README.md
```diff
@@ -7,7 +7,231 @@
...
+
+// A demuxed config.
+let result = AVCDecoderConfigurationRecord::demux(&mut io::Cursor::new(data.into())).unwrap();
+
+// Do something with it!
+
+// You can also access the sps bytestream and parse it:
+let sps = &result.sps[0];
+let sps = Sps::parse(sps.clone()).unwrap();
+```
```
TroyKomodo, line 208: I think, we only need an example on the demux where the data is raw bytes do not construct it via a bit writer.

crates/h264/README.md
```diff
@@ -7,7 +7,231 @@
...
+
+use bytes::Bytes;
+use scuffle_bytes_util::BitWriter;
+
+use scuffle_h264::{Sps, AVCDecoderConfigurationRecord};
+
+// Making a bytestream to demux
+let data = Bytes::from(b"\x01d\0\x1f\xff\xe1\0\x1dgd\0\x1f\xac\xd9A\xe0m\xf9\xe6\xa0  (\0\0\x03\0\x08\0\0\x03\x01\xe0x\xc1\x8c\xb0\x01\0\x06h\xeb\xe3\xcb\"\xc0\xfd\xf8\xf8\0".to_vec());
+// Demuxing
+let config = AVCDecoderConfigurationRecord::demux(&mut io::Cursor::new(data.clone())).unwrap();
```
TroyKomodo, line 225: In this case construct an avcdecoder config record from the struct.

crates/h264/README.md
```diff
@@ -7,7 +7,231 @@
...
+// 144 = time_scale / (2 * 100)
+// 28800 = time_scale
+// time_scale is a u32
+let _ = writer.write_bits(28800, 32);
+let _ = writer.finish();
+
+// Now result contains a complete SPS.
+let result = Sps::parse(sps.into()).unwrap();
+```
+For more SPS examples, check out the tests in the source code for SPS.
```
TroyKomodo, line 148: We dont want to use this lib to write the sps, this is only for parsing the sps.

crates/h264/README.md
```diff
@@ -7,7 +7,231 @@
...
+- less-avc
+  - This crate is focused on being a simple pure Rust implementation of H.264 but with only a subset of its features. IT lacks 4:4:4 support and some documentation for the specific SPS fields.
+- openh264-rs
+  - This crate is a set of bindings around the C implementation of H.264 (openh264). As a result, the crate contains a lot of unsafe code since it's just a lightweight rust wrapper.
+
+## Notable features
+
+This crate is a completely safe implementation of H264 encoding and decoding, which means there is no unsafe code!
+
+When combined with [scuffle-expgolomb](https://crates.io/crates/scuffle-expgolomb) and [scuffle-bytes-util](https://crates.io/crates/scuffle-bytes-util), working with h264 has never been easier!
```
TroyKomodo, line 33: The goal of this library is to be able to parse / work with the config to deal in the various containers. Maybe at a later point we can include a full encoder / decoder for h264 but for the moment it is only headers.

crates/h264/README.md
philipch07: Oh yeah that makes way more sense now

crates/h264/src/sps.rs
```diff
@@ -1,45 +1,812 @@
...
+
+    /// `constraint_set0_flag`: `1` if it abides by the constraints in A.2.1, `0` if unsure or otherwise.
+    ///
+    /// If `profile_idc` is 44, 100, 110, 122, or 244, this is automatically set to false.
+    ///
+    /// It is a single bit. ISO/IEC-14496-10-2022 - 7.4.2.1.1
+    ///
+    /// Note that this crate does NOT use the `constraint_set0_flag` field for decoding.
+    /// The parse function will still read the values if passed in correctly.
+    /// If you have an application that requires this feature, please reach out to us at <https://scuffle.cloud>.
```
TroyKomodo, line 57: We should just remove these notes

crates/h264/src/sps.rs
```diff
@@ -1,45 +1,812 @@
...
+    ///
+    /// `width = ((pic_width_in_mbs_minus1 + 1) * 16) - frame_crop_right_offset * 2 - frame_crop_left_offset * 2`
+    ///
+    /// This is a variable number of bits as it is encoded by an exp golomb (unsigned).
+    /// ISO/IEC-14496-10-2022 - 7.4.2.1.1
+    ///
+    /// For more information:
+    ///
+    /// <https://en.wikipedia.org/wiki/Exponential-Golomb_coding>
+    pub pic_width_in_mbs_minus1: u64,
```
TroyKomodo, line 297: We should create a method for doing this calculation `width()`

crates/h264/src/sps.rs
```diff
@@ -1,45 +1,812 @@
...
+    /// `height = ((2 - frame_mbs_only_flag as u64) * (pic_height_in_map_units_minus1 + 1) * 16) -
+    /// frame_crop_bottom_offset * 2 - frame_crop_top_offset * 2`
+    ///
+    /// This is a variable number of bits as it is encoded by an exp golomb (unsigned).
+    /// ISO/IEC-14496-10-2022 - 7.4.2.1.1
+    ///
+    /// For more information:
+    ///
+    /// <https://en.wikipedia.org/wiki/Exponential-Golomb_coding>
+    pub pic_height_in_map_units_minus1: u64,
```
TroyKomodo, line 312: We should create a method for doing this calculation `height()`

crates/h264/src/sps.rs
```diff
@@ -1,45 +1,812 @@
...
+    ///
+    /// `width = ((pic_width_in_mbs_minus1 + 1) * 16) - frame_crop_right_offset * 2 - frame_crop_left_offset * 2`
+    ///
+    /// This is a variable number of bits as it is encoded by an exp golomb (unsigned).
+    /// ISO/IEC-14496-10-2022 - 7.4.2.1.1
+    ///
+    /// For more information:
+    ///
+    /// <https://en.wikipedia.org/wiki/Exponential-Golomb_coding>
+    pub frame_crop_right_offset: u64,
```
TroyKomodo, line 383: Are these 2 fields only present if the `frame_cropping_flag` is set? if so, then we should make a new struct which is `FrameCroppingInfo` and make that an `Option`

crates/h264/src/sps.rs
```diff
@@ -1,45 +1,812 @@
...
+    /// `height = ((2 - frame_mbs_only_flag as u64) * (pic_height_in_map_units_minus1 + 1) * 16)
+    /// - frame_crop_bottom_offset * 2 - frame_crop_top_offset * 2`
+    ///
+    /// This is a variable number of bits as it is encoded by an exp golomb (unsigned).
+    /// ISO/IEC-14496-10-2022 - 7.4.2.1.1
+    ///
+    /// For more information:
+    ///
+    /// <https://en.wikipedia.org/wiki/Exponential-Golomb_coding>
+    pub frame_crop_bottom_offset: u64,
```
TroyKomodo, line 409: Didnt see this, but include these in the above.

crates/h264/src/sps.rs
```diff
@@ -1,45 +1,812 @@
...
+
+    /// The width as a u64. This is computed from other fields, and isn't directly set.
+    ///
+    /// `width = ((pic_width_in_mbs_minus1 + 1) * 16) - frame_crop_right_offset * 2 - frame_crop_left_offset * 2`
     pub width: u64,
+
+    /// The height as a u64. This is computed from other fields, and isn't directly set.
+    ///
+    /// `height = ((2 - frame_mbs_only_flag as u64) * (pic_height_in_map_units_minus1 + 1) * 16) - frame_crop_bottom_offset * 2 - frame_crop_top_offset * 2`
     pub height: u64,
```
TroyKomodo, line 419: We should remove this in favour of a method call.

TroyKomodo (commented):
I think in general, we should basically strive for the SPS to be a rust struct that represents the binary in a rust form. Instead of flags if things are present we just use `Option` and then we parse the flags (and potentially encode them back if we write)

crates/h264/src/sps.rs
philipch07: The 4 offsets default to 0 if the flag is unset. If the flag is set then we read in expgs for them.

Since I'll be slightly reformatting to structs so we dont need extra flags, I agree it would be good to use a `FrameCroppingInfo` struct instead.

crates/h264/src/sps.rs
```diff
@@ -1,45 +1,812 @@
...
 use byteorder::{BigEndian, ReadBytesExt};
 use bytes::Bytes;
 use scuffle_bytes_util::BitReader;
 use scuffle_expgolomb::BitReaderExpGolombExt;
 
+use crate::{AspectRatioIdc, NALUnitType, VideoFormat};
+
 #[derive(Debug, Clone, PartialEq)]
-/// Sequence parameter set
+/// The Sequence Parameter Set.
```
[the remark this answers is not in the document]
philipch07: I think at this point yes

crates/h264/src/sps.rs
```diff
@@ -53,24 +648,63 @@ impl Sps {
             return Err(io::Error::new(io::ErrorKind::InvalidData, "Forbidden zero bit is set"));
         }
 
-        bit_reader.seek_bits(2)?; // nal_ref_idc
-
-        let nal_unit_type = bit_reader.read_bits(5)?;
+        let nal_ref_idc = bit_reader.read_bits(2)? as u8;
+        let nal_unit_type = bit_reader.read_bits(5)? as u8;
         if nal_unit_type != 7 {
```
TroyKomodo, line 652: We should do the conversion to `NALUnitType` here and then check `NALUnitType::SPS` below

crates/h264/src/sps.rs
```diff
@@ -53,179 +647,412 @@ impl Sps {
...
+                offset_for_ref_frame.push(bit_reader.read_signed_exp_golomb()?);
             }
+
+            pic_order_cnt_type1 = Some(PicOrderCountType1 {
+                delta_pic_order_always_zero_flag,
+                offset_for_non_ref_pic,
+                offset_for_top_to_bottom_field,
+                num_ref_frames_in_pic_order_cnt_cycle,
+                offset_for_ref_frame,
+            })
```
TroyKomodo, line 740: We should move this into a `PicOrderCountType1::demux` function

crates/h264/src/sps.rs
```diff
@@ -53,179 +647,412 @@ impl Sps {
...
+            let frame_crop_top_offset = bit_reader.read_exp_golomb()?;
+            let frame_crop_bottom_offset = bit_reader.read_exp_golomb()?;
 
-        let vui_parameters_present_flag = bit_reader.read_bit()?;
+            frame_crop_info = Some(FrameCropInfo {
+                frame_crop_left_offset,
+                frame_crop_right_offset,
+                frame_crop_top_offset,
+                frame_crop_bottom_offset,
+            })
```
TroyKomodo, line 770: We should move this into a `FrameCropInfo::demux` function

crates/h264/src/sps.rs
```diff
@@ -53,179 +647,412 @@ impl Sps {
...
-                    bit_reader.seek_bits(16)?; // sar_width
-                    bit_reader.seek_bits(16)?; // sar_height
+                    sar_width = bit_reader.read_bits(16)? as u16;
+                    sar_height = bit_reader.read_bits(16)? as u16;
                 }
+
+                sample_aspect_ratio = Some(SarDimensions {
+                    aspect_ratio_idc: AspectRatioIdc(aspect_ratio_idc),
+                    sar_width,
+                    sar_height,
```
TroyKomodo, line 798: Likewise with the above

crates/h264/src/sps.rs
```diff
@@ -53,179 +647,412 @@ impl Sps {
...
+            base_width - (crop.frame_crop_left_offset + crop.frame_crop_right_offset) * 2
+        })
+    }
+
+    /// Returns the frame rate as a f64.
+    ///
+    /// If `timing_info_present_flag` is set, then the `frame_rate` will be computed, and
+    /// if `num_units_in_tick` is nonzero, then the framerate will be:
+    /// `frame_rate = time_scale as f64 / (2.0 * num_units_in_tick as f64)`
+    pub fn frame_rate(&self) -> f64 {
```
TroyKomodo, line 935: Should this be a `Option<f64>`, i am not sure. We are using 0.0 as a sentinel value (for not present) in rust which is an antipattern.

crates/h264/src/sps.rs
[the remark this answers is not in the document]
philipch07: Made a ticket: CLOUD-73

TroyKomodo (approved):

changes.d/pr-370.toml
```diff
@@ -0,0 +1,30 @@
+[[scuffle-h264]]
+category = "fix"
+description = "Fixes a bug with removing the emu prevention byte causing other bytes to be lost by accident"
+breaking = true
+authors = ["@philipch07"]
+
+[[scuffle-h264]]
+category = "refactor"
+description = "Refactored how tests work so they work like the tests in our other crates"
+breaking = true
```
TroyKomodo, line 10: this is non breaking

changes.d/pr-370.toml
```diff
@@ -0,0 +1,30 @@
+[[scuffle-h264]]
```
TroyKomodo, line 1: one more change log about the sps / config method / struct value changes

TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #372, opened 2025-02-20
chore: deny lints and improved docs

lennartkloock:
- Deny lints
- Improved docs

CLOUD-50

### DIFF
--- a/changes.d/pr-372.toml
+++ b/changes.d/pr-372.toml
@@ -1,0 +1,134 @@
+[[scuffle-aac]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-amf0]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-av1]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-batching]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-bootstrap]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-bootstrap-derive]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-bootstrap-telemetry]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-bytes-util]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-context]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-expgolomb]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-ffmpeg]]
+category = "chore"
+description = "deny lints"
+authors = ["@lennartkloock"]
+
+[[scuffle-flv]]
+category = "chore"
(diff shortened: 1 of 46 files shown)


crates/aac/src/lib.rs
```diff
@@ -8,6 +8,9 @@
 //!
 //! `SPDX-License-Identifier: MIT OR Apache-2.0`
 #![cfg_attr(all(coverage_nightly, test), feature(coverage_attribute))]
+#![deny(missing_docs)]
+#![deny(unsafe_code)]
+#![deny(clippy::missing_const_for_fn)]
```
TroyKomodo, line 13: let me change the ticket to get rid of this one. This one is a double edged sword. Lets not use const fn unless something we know is const. I mentioned this in the other pr #370 

> ive been thinking abt this lint. its a double edged sword, on one point its nice to have functions that are const fn since therefore they can be used in const contexts. On the other hand it means we are locking ourselves into this if we change the fn implementation and makes it so we cant use a const fn it is a breaking change. I think we should remove this and avoid using const fn unless we can 100% justify that the function can never be non-const.

crates/mp4/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![deny(unsafe_code)]
```
TroyKomodo, line 1: I know a future pr will address mp4 so perhaps let's remove this for now and address it then.

crates/nutype_enum/src/lib.rs
```diff
@@ -47,6 +75,7 @@ macro_rules! nutype_enum {
             $(
                 $(#[$variant_attr])*
                 #[allow(non_upper_case_globals)]
+                #[allow(missing_docs)]
```
TroyKomodo, line 78: This change is wrong. We should still force variants to require docs if the crate has a requirement for docs

crates/rtmp/src/lib.rs
```diff
@@ -1,3 +1,19 @@
+//! A crate for handling RTMP server connections.
```
TroyKomodo, line 1: Let's remove this and let the rtmp pr handle this

crates/batching/src/batch.rs
```diff
@@ -1,3 +1,5 @@
+//! Types related to the batcher and batch executor.
```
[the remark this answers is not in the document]
TroyKomodo: just maybe a doc example, followed by the difference between batching / dataloading.

crates/flv/src/lib.rs
```diff
@@ -19,6 +19,7 @@
 //!
 //! `SPDX-License-Identifier: MIT OR Apache-2.0`
 #![cfg_attr(all(coverage_nightly, test), feature(coverage_attribute))]
+#![deny(unsafe_code)]
```
philipch07, line 23: Is this missing the missing docs flag? If we're intentionally excluding the flag then we should make a ticket for the flv crate if necessary.

lennartkloock: Yes, there is a ticket for that. CLOUD-66

TroyKomodo (approved):

changes.d/pr-372.toml
```diff
@@ -0,0 +1,109 @@
...
+
+[[scuffle-ffmpeg]]
+category = "docs"
+description = "improved documentation"
+authors = ["@lennartkloock"]
+
+[[scuffle-flv]]
+category = "docs"
+description = "improved documentation"
+authors = ["@lennartkloock"]
```
philipch07, line 59: this pr doesn't change flv i think

philipch07 (approved):

lennartkloock:
?brawl merge

PR #377, opened 2025-02-22
add ignore for RUSTSEC-2025-0007

TroyKomodo:
ring is unmaintained
we do not use ring for anything except for http3 testing in scuffle-http. Unfortunately reqwest uses ring for http3.
See https://github.com/seanmonstar/reqwest/issues/2566


philipch07 (approved):

TroyKomodo:
?brawl merge

PR #384, opened 2025-02-25
Adds builders for h264 sps structs + size of un/signed Exp-Golomb encoded numbers

philipch07:
Since we have parsers for each `Sps` struct (including substructs) we should be able to also rebuild them. This enables us to directly use `Sps`s in the `sps` vec in the config.

This also adds the ability to get the number of bits that an un/signed expg will take without having to write it.

Check the associated change-log for more info.

CLOUD-76


crates/h264/src/config.rs
```diff
@@ -227,8 +234,8 @@ impl AVCDecoderConfigurationRecord {
 
             bit_writer.write_bits(config.sequence_parameter_set_ext.len() as u64, 8)?;
             for sps_ext in &config.sequence_parameter_set_ext {
-                bit_writer.write_u16::<BigEndian>(sps_ext.len() as u16)?;
-                bit_writer.write_all(sps_ext)?;
+                bit_writer.write_u16::<BigEndian>(sps_ext.bytesize() as u16)?;
+                sps_ext.build(&mut bit_writer)?;
```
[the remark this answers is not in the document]
philipch07: Honestly I'm not sure. The only reason why I added the align for the sps section and not the ext section was because there was a test case that would cause the rebuilt config to be different if not aligned, but I didn't encounter any issues when rebuilding the ext, probably because there aren't any reductions that happen when writing the ext.

crates/h264/src/sps.rs
```diff
@@ -549,6 +550,198 @@ impl Sps {
         })
     }
 
+    /// Builds the Sps struct into a byte stream.
+    /// Returns a built byte stream.
+    pub fn build<T: io::Write>(&self, writer: &mut BitWriter<T>) -> io::Result<()> {
```
TroyKomodo, line 555: This should not take in a bitwriter. It should just be a `std::io::Write` we should then construct it using a bitwriter and then add the emulation prevention byte when 3 zero bytes are together.

crates/h264/src/sps.rs
```diff
@@ -2120,36 +2010,13 @@ mod tests {
...
-        let mut reduced_buf = Vec::new();
-        let mut writer3 = BitWriter::new(&mut reduced_buf);
-
-        reduced.build(&mut writer3).unwrap();
-        writer3.finish().unwrap();
-        assert_eq!(reduced_buf, buf);
-
-        // now we can check the size:
-        assert_eq!(reduced.size(), result.size());
+        result.build(&mut buf).unwrap();
```
TroyKomodo, line 2019: we should check the buffer is the same as the input sps

crates/h264/src/sps.rs
```diff
@@ -2770,30 +2631,14 @@ mod tests {
...
-        // by rebuilding it.
-        let reduced = Sps::parse(&buf).unwrap();
-        assert_eq!(reduced, result);
+        let rebuilt = Sps::parse(std::io::Cursor::new(&buf)).unwrap();
 
-        // now we can check that the bitstream from
-        // the reduced version should be the same
-        let mut reduced_buf = Vec::new();
-        let mut writer3 = BitWriter::new(&mut reduced_buf);
+        dbg!(&rebuilt);
```
TroyKomodo, line 2639: stray dbg

crates/h264/src/lib.rs
```diff
@@ -25,18 +25,18 @@
...
+//! let data = Bytes::from(b"\x01d\0\x1f\xff\xe1\0\x17\x67\x64\x00\x1F\xAC\xD9\x41\xE0\x6D\xF9\xE6\xA0\x20\x20\x28\x00\x00\x00\x08\x00\x00\x01\xE0\x01\0\x06h\xeb\xe3\xcb\"\xc0\xfd\xf8\xf8\0".to_vec());
 //!
 //! // Parsing
 //! let result = AVCDecoderConfigurationRecord::parse(&mut io::Cursor::new(data.into())).unwrap();
 //!
 //! // Do something with it!
 //!
-//! // You can also access the sps bytestream and parse it:
-//! let sps = Sps::parse(&result.sps[0]).unwrap();
+//! // You can also access the contained Sps struct:
```
TroyKomodo, line 38: Update docs

crates/h264/src/lib.rs
```diff
@@ -25,18 +25,18 @@
 //!
 //! use bytes::Bytes;
 //!
-//! use scuffle_h264::{Sps, AVCDecoderConfigurationRecord};
+//! use scuffle_h264::{AVCDecoderConfigurationRecord, Sps};
 //!
 //! // A sample h264 bytestream to parse
-//! let data = Bytes::from(b"\x01d\0\x1f\xff\xe1\0\x1dgd\0\x1f\xac\xd9A\xe0m\xf9\xe6\xa0  (\0\0\x03\0\x08\0\0\x03\x01\xe0x\xc1\x8c\xb0\x01\0\x06h\xeb\xe3\xcb\"\xc0\xfd\xf8\xf8\0".to_vec());
+//! let data = Bytes::from(b"\x01d\0\x1f\xff\xe1\0\x17\x67\x64\x00\x1F\xAC\xD9\x41\xE0\x6D\xF9\xE6\xA0\x20\x20\x28\x00\x00\x00\x08\x00\x00\x01\xE0\x01\0\x06h\xeb\xe3\xcb\"\xc0\xfd\xf8\xf8\0".to_vec());
```
TroyKomodo, line 31: Maybe hide this in the docs since raw bytes dont make a lot of sense. Rename the variable to `bytes`  and then prefix the line with a `#` to disable it from being rendered by the docs.rs

crates/h264/src/sps.rs
```diff
@@ -340,35 +340,11 @@ pub struct Sps {
 }
 
 impl Sps {
-    /// Parsees an SPS from the input bytes.
+    /// Parses an Sps from the input bytes.
```
TroyKomodo, line 343: We should split this file into a module. 

and then each sub type has its own file.

crates/flv/src/lib.rs
```diff
@@ -237,63 +236,9 @@ mod tests {
...
-                chroma_sample_loc: None,
-                timing_info: Some(
-                    TimingInfo {
-                        num_units_in_tick: 1,
-                        time_scale: 120,
-                    },
-                ),
-            }
-            ");
+            let sps = &avc_decoder_configuration_record.sps[0];
```
TroyKomodo, line 239: we should undo these changes and just parse it

changes.d/pr-384.toml
```diff
@@ -0,0 +1,53 @@
...
+category = "feat"
+description = "Building an Sps will automatically insert necessary emulation prevention bytes (which are already properly handled in the existing parse function)"
+breaking = false
+authors = ["@TroyKomodo"]
+
+[[scuffle-h264]]
+category = "feat"
+description = "As a result of being able to build, we also have the ability to generate minimal and functionally equivalent Sps headers via `Sps::reduce()`"
+breaking = false
+authors = ["@philipch07"]
```
TroyKomodo, line 41: These can be removed, and replaced with something talking about the `EmulationBytesIo` thingy

TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #392, opened 2025-03-02
fix missed test case in h264 pr

TroyKomodo:
In the test we incorrectly use `Sps::parse` and instead should be using `Sps::parse_with_emulation_prevention`. Also corrects the README.md docs to do the same.


lennartkloock (approved):

lennartkloock:
?brawl merge

PR #394, opened 2025-03-02
feat(http): onion service backend using arti

lennartkloock:
- Support for running `scuffle-http` with an onion service backend using the `arti` crate


TroyKomodo:
@lennartkloock why again did we not merge this i forgot

lennartkloock:
> @lennartkloock why again did we not merge this i forgot

Because it relies on libsqlite and some other native libs to be installed to compile.

PR #395, opened 2025-03-03
Windows support for scuffle-signal

SimaoMoreira5228:
this pr aims to add windows upport for the signal crate

### DIFF
--- a/crates/signal/src/lib.rs
+++ b/crates/signal/src/lib.rs
@@ -244,7 +244,7 @@ impl SignalKind {
         if cfg!(test) {
             return Ok(WindowsSignalValue::Mock(
                 *self,
-                Box::pin(tokio_stream::wrappers::BroadcastStream::new(SignalMocker::subscribe())),
+                Box::pin(tokio_stream::wrappers::BroadcastStream::new(test::SignalMocker::subscribe())),
             ));
         }
 
@@ -408,7 +408,7 @@ mod test {
     use crate::{SignalHandler, SignalKind};
 
     #[cfg(windows)]
-    struct SignalMocker(tokio::sync::broadcast::Sender<SignalKind>);
+    pub(crate) struct SignalMocker(tokio::sync::broadcast::Sender<SignalKind>);
 
     #[cfg(windows)]
     impl SignalMocker {
@@ -423,7 +423,7 @@ mod test {
             SIGNAL_MOCKER.with(|local| local.0.send(kind).unwrap());
         }
 
-        fn subscribe() -> tokio::sync::broadcast::Receiver<SignalKind> {
+        pub(crate) fn subscribe() -> tokio::sync::broadcast::Receiver<SignalKind> {
             println!("subscribing");
             SIGNAL_MOCKER.with(|local| local.0.subscribe())
         }
@@ -436,7 +436,7 @@ mod test {
 
     #[cfg(windows)]
     pub async fn raise_signal(kind: SignalKind) {
-        crate::SignalMocker::raise(kind);
+        SignalMocker::raise(kind);
     }
 
     #[cfg(unix)]


crates/signal/src/lib.rs
```diff
@@ -65,15 +63,128 @@
 use std::pin::Pin;
```
TroyKomodo, line 63: There is a cfg atter above this which prevents this crate from being compiled on windows, we should remove that.

crates/signal/src/lib.rs
```diff
@@ -65,15 +63,128 @@
...
+    fn eq(&self, other: &UnixSignalKind) -> bool {
+        match self {
+            Self::Unix(kind) => kind == other,
+            _ => false,
+        }
+    }
+}
+
+#[cfg(windows)]
+enum WindowsSignalKind {
```
TroyKomodo, line 105: I think this enum should be public, so people can pass the values directly

crates/signal/src/lib.rs
```diff
@@ -65,15 +63,128 @@
...
+
+#[cfg(unix)]
+impl PartialEq<UnixSignalKind> for SignalKind {
+    fn eq(&self, other: &UnixSignalKind) -> bool {
+        match self {
+            Self::Unix(kind) => kind == other,
+            _ => false,
+        }
+    }
+}
```
TroyKomodo, line 102: i think we should also take into account, partial equals for the interrupt / term being compared to unix interrupt / term.

crates/signal/src/lib.rs
```diff
@@ -65,15 +63,128 @@
...
+    Self::Windows(value)
+  }
+}
+
+#[cfg(windows)]
+impl PartialEq<WindowsSignalKind> for SignalKind {
+    fn eq(&self, other: &WindowsSignalKind) -> bool {
+        match self {
+            Self::Windows(kind) => kind == other,
+            _ => false,
```
TroyKomodo, line 134: likewise with the above

crates/signal/src/lib.rs
```diff
@@ -65,15 +63,128 @@
...
+            Self::Interrupt => Ok(SignalValue::CtrlC(tokio::signal::windows::ctrl_c()?)),
+            // https://learn.microsoft.com/en-us/windows/console/ctrl-close-signal
+            Self::Terminate => Ok(SignalValue::CtrlClose(tokio::signal::windows::ctrl_close()?)),
+            Self::Windows(kind) => match kind {
+                WindowsSignalKind::CtrlBreak => Ok(SignalValue::CtrlBreak(tokio::signal::windows::ctrl_break()?)),
+                WindowsSignalKind::CtrlC => Ok(SignalValue::CtrlC(tokio::signal::windows::ctrl_c()?)),
+                WindowsSignalKind::CtrlClose => Ok(SignalValue::CtrlClose(tokio::signal::windows::ctrl_close()?)),
+                WindowsSignalKind::CtrlLogoff => Ok(SignalValue::CtrlLogoff(tokio::signal::windows::ctrl_logoff()?)),
+                WindowsSignalKind::CtrlShutdown => Ok(SignalValue::CtrlShutdown(tokio::signal::windows::ctrl_shutdown()?)),
+            },
```
TroyKomodo, line 181: this match statement can be collapsed into a single match

crates/signal/src/lib.rs
```diff
@@ -213,7 +326,7 @@ mod tests {
 
     use super::*;
 
-    pub fn raise_signal(kind: SignalKind) {
+    pub fn raise_signal(kind: UnixSignalKind) {
```
TroyKomodo, line 329: we should try make this work for both windows / unix

TroyKomodo:
We don't allow merge commits to be merged, so you should rebase this pr.

crates/signal/src/lib.rs
```diff
@@ -1,8 +1,6 @@
 //! A crate designed to provide a more user friendly interface to
 //! `tokio::signal`.
 //!
-//! Currently only supported on Unix systems. (Windows is planned but not yet implemented)
-//!
```
TroyKomodo, line 5: There needs to be an update in the README.md too.

crates/signal/src/bootstrap.rs
```diff
@@ -13,10 +13,10 @@ pub trait SignalConfig: Global {
     /// The signals to listen for.
     ///
     /// By default, listens for `SIGTERM` and `SIGINT`.
-    fn signals(&self) -> Vec<tokio::signal::unix::SignalKind> {
+    fn signals(&self) -> Vec<crate::SignalKind> {
         vec![
-            tokio::signal::unix::SignalKind::terminate(),
-            tokio::signal::unix::SignalKind::interrupt(),
+            crate::SignalKind::Unix(tokio::signal::unix::SignalKind::terminate()),
+            crate::SignalKind::Unix(tokio::signal::unix::SignalKind::interrupt()),
```
TroyKomodo, line 19: We should change these to be the crate::SignalKind::Interupt / Terminate ones which are generic across os.

crates/signal/src/lib.rs
```diff
@@ -65,15 +63,128 @@
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-use tokio::signal::unix::{Signal, SignalKind};
+#[cfg(unix)]
+use tokio::signal::unix::{Signal as UnixSignal, SignalKind as UnixSignalKind};
```
TroyKomodo, line 67: we should just import this as unix to be the same with windows; and then we should reexport the `unix::UnixSignalKind`

crates/signal/Cargo.toml
```diff
@@ -19,13 +19,14 @@ scuffle-bootstrap = { workspace = true, optional = true }
 scuffle-context = { workspace = true, optional = true }
 anyhow = { version = "1", optional = true }
 scuffle-workspace-hack.workspace = true
+libc = "0.2"
```
TroyKomodo, line 22: Why is this no longer a dev dep?

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,164 @@
...
 
-/// A handler for listening to multiple Unix signals, and providing a future for
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SignalKind {
+    Interrupt,
+    Terminate,
+    #[cfg(windows)]
+    Windows(WindowsSignalKind),
+    #[cfg(unix)]
+    Unix(unix::SignalKind),
```
TroyKomodo, line 85: We import this as UnixSignalKind so we can just use that here

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,164 @@
...
+}
+
+#[cfg(unix)]
+impl From<unix::SignalKind> for SignalKind {
```
TroyKomodo, line 89: Same here

crates/signal/src/lib.rs
```diff
@@ -215,35 +359,43 @@ mod tests {
 
     pub fn raise_signal(kind: SignalKind) {
```
TroyKomodo, line 360: We should just make 2 versions of this function I think

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,164 @@
...
+                Ok(SignalValue::CtrlLogoff(tokio::signal::windows::ctrl_logoff()?))
+            }
+            Self::Windows(WindowsSignalKind::CtrlShutdown) => {
+                Ok(SignalValue::CtrlShutdown(tokio::signal::windows::ctrl_shutdown()?))
+            }
+        }
+    }
+
+    #[cfg(unix)]
+    fn as_raw_value(&self) -> i32 {
```
TroyKomodo, line 193: Are these functions only used in tests? If so we can just do #[cfg(test)] on them

crates/signal/Cargo.toml
SimaoMoreira5228: image
i used i there, but i can use static values as i did on windows if its better

crates/signal/src/lib.rs
SimaoMoreira5228: should i change all the appearances of unix::SignalKind in this file?

crates/signal/src/lib.rs
SimaoMoreira5228: for now they are only use in tests, but i don't know if they should be made public

crates/signal/src/lib.rs
TroyKomodo: `UnixSignalKind` we import the type under an alias

crates/signal/src/lib.rs
TroyKomodo: maybe later, for now u can just cfg test them

crates/signal/Cargo.toml
SimaoMoreira5228: or should i put the winapi on dependencies and use it instead of static values?

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,166 @@
...
+            }
+            Self::Windows(WindowsSignalKind::CtrlShutdown) => {
+                Ok(SignalValue::CtrlShutdown(tokio::signal::windows::ctrl_shutdown()?))
+            }
+        }
+    }
+
+    #[cfg(unix)]
+    #[cfg(test)]
+    fn as_raw_value(&self) -> i32 {
```
TroyKomodo, line 194: maybe better to do `#[cfg(all(unix, test))]`

SimaoMoreira5228: thats cool

crates/signal/src/lib.rs
```diff
@@ -213,37 +359,47 @@ mod tests {
...
 
-        raise_signal(SignalKind::user_defined2());
+        raise_signal(SignalKind::Unix(UnixSignalKind::user_defined2()));
 
         // We should be able to receive the signal again
         let recv = (&mut handler).with_timeout(Duration::from_millis(5)).await.unwrap();
 
-        assert_eq!(recv, SignalKind::user_defined2(), "expected SIGUSR2");
+        assert_eq!(recv, UnixSignalKind::user_defined2(), "expected SIGUSR2");
     }
```
TroyKomodo, line 403: these tests will not work on windows, we should disable them on windows and add a new windows only test.

crates/signal/Cargo.toml
TroyKomodo: We can move this back now since its only use on tests

crates/signal/Cargo.toml
```diff
@@ -23,9 +23,10 @@ scuffle-workspace-hack.workspace = true
 [dev-dependencies]
 tokio = { version = "1.41.1", features = ["macros", "rt", "time"] }
 tokio-test = "0.4.4"
-libc = "0.2"
 futures = "0.3"
 scuffle-future-ext.workspace = true
+libc = "0.2"
+winapi = "0.3.9"
```
TroyKomodo, line 29: this should be a windows only dep

```
[target.'cfg(windows)'.dev-dependencies]
winapi = "0.3.9"
```

crates/signal/src/lib.rs
```diff
@@ -213,60 +357,70 @@ mod tests {
...
+    #[cfg(windows)]
+    pub fn raise_signal(kind: SignalKind) {
+        unsafe {
+            use winapi::um::winconGenerateConsoleCtrlEvent;
+
+            GenerateConsoleCtrlEvent(kind.as_raw_value(), 0);
+        }
+    }
+
+    #[cfg(all(not(valgrind), unix))] // test is time-sensitive
```
TroyKomodo, line 377: we should then make a new test like this but specifically for windows

crates/signal/src/lib.rs
```diff
@@ -213,60 +357,70 @@ mod tests {
...
         // Safety: This is a test, and we control the process.
         unsafe {
             libc::raise(kind.as_raw_value());
         }
     }
 
-    #[cfg(not(valgrind))] // test is time-sensitive
+    #[cfg(windows)]
+    pub fn raise_signal(kind: SignalKind) {
+        unsafe {
```
TroyKomodo, line 370: needs a safety comment, about why `GenerateConsoleControlEvent` is okay to call

crates/signal/Cargo.toml
SimaoMoreira5228: i don't think it is used for anything windows related

TroyKomodo: sure it can be

SimaoMoreira5228: is it even possible using `libc` on windows aside from `Cygwin` or `MSYS2`?

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,164 @@
...
+    #[cfg(windows)]
+    Windows(WindowsSignalKind),
+    #[cfg(unix)]
+    Unix(UnixSignalKind),
+}
+
+#[cfg(unix)]
+impl From<UnixSignalKind> for SignalKind {
+    fn from(value: UnixSignalKind) -> Self {
+        Self::Unix(value)
```
TroyKomodo, line 91: I think on this conversion we check if its an interrupt or terminate signal and if it is use the predefined type.

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,164 @@
...
+    CtrlC(tokio::signal::windows::CtrlC),
+    CtrlClose(tokio::signal::windows::CtrlClose),
+    CtrlLogoff(tokio::signal::windows::CtrlLogoff),
+    CtrlShutdown(tokio::signal::windows::CtrlShutdown),
+}
+
+#[cfg(windows)]
+impl From<WindowsSignalKind> for SignalKind {
+    fn from(value: WindowsSignalKind) -> Self {
+        Self::Windows(value)
```
TroyKomodo, line 127: likewise for the unix one, check if its interrupt / term

crates/signal/src/lib.rs
```diff
@@ -76,6 +76,11 @@ mod bootstrap;
 pub use bootstrap::{SignalConfig, SignalSvc};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// The kind of signal to listen for.
+/// The `Interrupt` maps to `SIGINT` on Unix, and `Ctrl-C` on Windows.
+/// The `Terminate` maps to `SIGTERM` on Unix, and `Ctrl-Close` on Windows.
+/// The `Unix` and `Windows` are plataform specific signals. `Unix` maps to
+/// `UnixSignalKind`, and `Windows` maps to `WindowsSignalKind`.
```
TroyKomodo, line 83: This comment is on the struct, we should just say something like
```
The type of signal to listen for. 
``` 

and then we should put the stuff about `Interrupt` above the variant.

Comments should also go above the derive statements.

crates/signal/src/lib.rs
```diff
@@ -213,60 +378,127 @@ mod tests {
...
+            use winapi::um::wincon::GenerateConsoleCtrlEvent;
+
+            match kind {
+                SignalKind::Interrupt => GenerateConsoleCtrlEvent(winapi::um::wincon::CTRL_C_EVENT, 0),
+                SignalKind::Windows(WindowsSignalKind::CtrlBreak) => {
+                    GenerateConsoleCtrlEvent(winapi::um::wincon::CTRL_BREAK_EVENT, 0)
+                }
+                SignalKind::Windows(WindowsSignalKind::CtrlC) => {
+                    GenerateConsoleCtrlEvent(winapi::um::wincon::CTRL_C_EVENT, 0)
+                }
```
TroyKomodo, line 403: we should be doing it like `SignalKind::Terminate | SignalKind::Windows(WindowsSignalKind::CtrlBreak)`

SimaoMoreira5228: SignalKind::Terminate corresponds to [CTRL+CLOSE Signal](https://learn.microsoft.com/en-us/windows/console/ctrl-close-signal), but i will do that to Interrupt and CtrlC signals

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

TroyKomodo:
?brawl try

changes.d/pr-395.toml
```diff
@@ -0,0 +1,5 @@
+[[scuffle-signal]]
+category = "feat"
+description = "add windows support"
+breaking = true
+authors = ["@SimaoMoreira5228"]
```
TroyKomodo, line 5: we should also talk abt the new function on the signal config for bootstrap

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,246 @@
...
+    Terminate,
+    /// Represents a Windows-specific signal kind, as defined in `WindowsSignalKind`.
+    #[cfg(windows)]
+    Windows(WindowsSignalKind),
+    /// Represents a Unix-specific signal kind, wrapping `tokio::signal::unix::SignalKind`.
+    #[cfg(unix)]
+    Unix(UnixSignalKind),
+}
+
+impl PartialEq for SignalKind {
```
TroyKomodo, line 91: maybe we should add a unit test for this

crates/signal/Cargo.toml
```diff
@@ -21,11 +21,17 @@ anyhow = { version = "1", optional = true }
...
 [dev-dependencies]
-tokio = { version = "1.41.1", features = ["macros", "rt", "time"] }
-tokio-test = "0.4.4"
-libc = "0.2"
+tokio = { version = "1.41.1", features = ["full"] }
+tokio-test = "0.4"
 futures = "0.3"
 scuffle-future-ext.workspace = true
+libc = "0.2"
+libtest-mimic = "0.8"
```
TroyKomodo, line 29: this isnt needed anymore

crates/signal/Cargo.toml
```diff
@@ -21,11 +21,17 @@ anyhow = { version = "1", optional = true }
...
+
+[target.'cfg(windows)'.dev-dependencies]
+windows = { version = "0.60.0", features = ["Win32_Storage_FileSystem", "Win32_System_Console", "Win32_System_Threading", "Win32_Security", "Win32_System_IO"] }
```
TroyKomodo, line 32: not sure if we use this

crates/signal/Cargo.toml
```diff
@@ -21,11 +21,17 @@ anyhow = { version = "1", optional = true }
...
+windows-strings = "0.3"
```
TroyKomodo, line 33: or this

crates/signal/README.md
```diff
@@ -9,8 +9,6 @@
 
 A crate designed to provide a more user friendly interface to `tokio::signal`.
 
-Currently only supported on Unix systems. (Windows is planned but not yet implemented)
```
TroyKomodo, line 12: We should update the readme example to be the same as the module example

crates/signal/src/bootstrap.rs
```diff
@@ -43,6 +40,11 @@ pub trait SignalConfig: Global {
 
         std::future::ready(Err(err))
     }
+
+    /// The global shutdown future.
+    fn block_global_shutdown(&self) -> impl std::future::Future<Output = ()> + Send {
+        scuffle_context::Handler::global().shutdown()
+    }
```
TroyKomodo, line 47: Maybe add some more documentation  as to what this does exactly.

So in this case this function allows the user to override the context that is shutdown when the signal is canceled.

TroyKomodo (commented):
@philipch07 @lennartkloock can you check this out

SimaoMoreira5228:
@philipch07 check it out when you can

crates/signal/src/bootstrap.rs
```diff
@@ -82,24 +88,18 @@ impl<Global: SignalConfig> Service<Global> for SignalSvc {
     }
 }
 
-#[cfg_attr(all(coverage_nightly, test), coverage(off))]
```
philipch07, line 85: Why did we remove this?

crates/signal/src/bootstrap.rs
```diff
@@ -82,24 +88,18 @@ impl<Global: SignalConfig> Service<Global> for SignalSvc {
...
-    use tokio::signal::unix::SignalKind;
 
-    use super::{SignalConfig, SignalSvc};
-    use crate::SignalHandler;
-    use crate::tests::raise_signal;
+    use super::*;
+    use crate::SignalKind;
+    use crate::test::raise_signal;
 
     async fn force_shutdown_two_signals<Global: GlobalWithoutConfig + SignalConfig>() {
```
philipch07, line 100: We should flag all tests that are time sensitive with the flag that prevents valgrind from running on them + leave a comment saying that the test is time sensitive. Unfortunately I don't exactly remember where or what the flag is but I know it exists somewhere.

crates/signal/src/bootstrap.rs
```diff
@@ -139,45 +140,66 @@ mod tests {
...
         }
+
+        // We dont want to block the global shutdown
+        async fn block_global_shutdown(&self) {
+            self.0.notified().await;
+        }
     }
 
     #[tokio::test]
     async fn bootstrap_service_no_timeout() {
```
philipch07, line 174: Same as above

crates/signal/src/bootstrap.rs
```diff
@@ -197,43 +219,34 @@ mod tests {
...
             None
         }
+
+        async fn block_global_shutdown(&self) {
+            std::future::pending().await
+        }
     }
 
     #[tokio::test]
     async fn bootstrap_service_no_signals() {
```
philipch07, line 236: Same as above

crates/signal/src/bootstrap.rs
```diff
@@ -249,7 +262,11 @@ mod tests {
 
     impl SignalConfig for SmallTimeoutTestGlobal {
```
philipch07, line 263: Same as above

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,246 @@
...
+    Interrupt,
+    /// Represents the terminate signal, which is `SIGTERM` on Unix and `Ctrl-Close` on Windows.
+    Terminate,
+    /// Represents a Windows-specific signal kind, as defined in `WindowsSignalKind`.
+    #[cfg(windows)]
+    Windows(WindowsSignalKind),
+    /// Represents a Unix-specific signal kind, wrapping `tokio::signal::unix::SignalKind`.
+    #[cfg(unix)]
+    Unix(UnixSignalKind),
+}
```
philipch07, line 89: @TroyKomodo do you think this should be a nutype enum or not since it doesn't have to be tied to a specific value?

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,246 @@
...
+            #[cfg(windows)]
+            (
+                Self::Interrupt | Self::Windows(WindowsSignalKind::CtrlC),
+                Self::Interrupt | Self::Windows(WindowsSignalKind::CtrlC),
+            ) => true,
+            #[cfg(windows)]
+            (
+                Self::Terminate | Self::Windows(WindowsSignalKind::CtrlClose),
+                Self::Terminate | Self::Windows(WindowsSignalKind::CtrlClose),
+            ) => true,
```
philipch07, line 108: What's the reasoning behind why these are here twice?

crates/signal/src/lib.rs
```diff
@@ -205,68 +432,134 @@ impl std::future::Future for SignalHandler {
...
+                SignalKind::Terminate => libc::SIGTERM,
+                SignalKind::Unix(kind) => kind.as_raw_value(),
+            });
         }
     }
 
-    #[cfg(not(valgrind))] // test is time-sensitive
+    #[cfg(windows)]
+    #[tokio::test]
+    async fn signal_handler() {
```
philipch07, line 462: Might want the valgrind flag again

crates/signal/src/lib.rs
```diff
@@ -205,68 +432,134 @@ impl std::future::Future for SignalHandler {
...
+        raise_signal(SignalKind::Windows(WindowsSignalKind::CtrlBreak)).await;
+
+        let recv = (&mut handler).with_timeout(Duration::from_millis(500)).await.unwrap();
+
+        assert_eq!(recv, WindowsSignalKind::CtrlBreak, "expected CtrlBreak");
+    }
+
+    #[cfg(windows)]
+    #[tokio::test]
+    async fn add_signal() {
```
philipch07, line 487: Same as above

crates/signal/src/lib.rs
```diff
@@ -205,68 +432,134 @@ impl std::future::Future for SignalHandler {
...
+        raise_signal(SignalKind::Windows(WindowsSignalKind::CtrlBreak)).await;
+
+        let recv = handler.recv().with_timeout(Duration::from_millis(500)).await.unwrap();
+
+        assert_eq!(recv, WindowsSignalKind::CtrlBreak, "expected CtrlBreak");
+    }
+
+    #[cfg(all(not(valgrind), unix))] // test is time-sensitive
     #[tokio::test]
     async fn signal_handler() {
```
philipch07, line 514: same as above

crates/signal/src/lib.rs
```diff
@@ -205,68 +432,134 @@ impl std::future::Future for SignalHandler {
...
+        let recv = (&mut handler).with_timeout(Duration::from_millis(500)).await.unwrap();
 
-        assert_eq!(recv, SignalKind::user_defined2(), "expected SIGUSR2");
+        assert_eq!(recv, UnixSignalKind::user_defined2(), "expected SIGUSR2");
     }
 
-    #[cfg(not(valgrind))] // test is time-sensitive
+    #[cfg(all(not(valgrind), unix))] // test is time-sensitive
     #[tokio::test]
     async fn add_signal() {
```
philipch07, line 542: same as above

crates/signal/src/bootstrap.rs
philipch07: Nvm for windows tests we cannot run valgrind anyways.

The only tests that should have the valgrind flags are the tests that do NOT have the windows flag AND are time sensitive.

crates/signal/src/lib.rs
philipch07: Nvm ignore this

crates/signal/src/bootstrap.rs
```diff
@@ -82,24 +88,19 @@ impl<Global: SignalConfig> Service<Global> for SignalSvc {
...
+#[cfg_attr(all(coverage_nightly, test), coverage(off))]
+mod test {
+    use scuffle_bootstrap::GlobalWithoutConfig;
     use scuffle_future_ext::FutureExt;
-    use tokio::signal::unix::SignalKind;
 
-    use super::{SignalConfig, SignalSvc};
-    use crate::SignalHandler;
-    use crate::tests::raise_signal;
+    use super::*;
```
philipch07, line 97: Can we try to avoid using the wildcard and instead manually grab the required things?

philipch07 (approved):

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,246 @@
...
+            },
+        }
+    }
+}
+
+#[cfg(unix)]
+type Signal = unix::Signal;
+
+#[cfg(windows)]
+type Signal = WindowsSignalValue;
```
lennartkloock, line 256: Is this is a good idea? I think the same name shouldn't refer to different types depending on the compile target. Why is this needed?

crates/signal/src/lib.rs
```diff
@@ -149,27 +376,27 @@ impl SignalHandler {
...
     /// Add a signal to the handler.
     ///
     /// If the signal is already in the handler, it will not be added again.
-    pub fn add_signal(&mut self, kind: SignalKind) -> &mut Self {
+    pub fn add_signal(&mut self, kind: impl Into<SignalKind>) -> &mut Self {
+        // Windows handles signals differently from unix.
+        // Windows signals are sent to a "console". Any process that is attached to the console will receive the signal.
+        // It happens that the test harness is attached to the same console as the test process, meaning that
+        // when we raise a signal it will be received by the harness and the test will cancel.
+        // This is a hack to get around that.
```
lennartkloock, line 392: Why is this comment here? What does it have to do with this function?

lennartkloock (changes requested):

crates/signal/src/lib.rs
```diff
@@ -60,20 +58,246 @@
...
+            Self::Interrupt => tokio::signal::unix::signal(UnixSignalKind::interrupt()),
+            Self::Terminate => tokio::signal::unix::signal(UnixSignalKind::terminate()),
+            Self::Unix(kind) => tokio::signal::unix::signal(*kind),
+        }
+    }
+
+    #[cfg(windows)]
+    fn listen(&self) -> Result<Signal, std::io::Error> {
+        #[cfg(test)]
+        if cfg!(test) {
```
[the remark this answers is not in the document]
TroyKomodo: If we remove it clippy will complain about being unreachable

crates/signal/src/lib.rs
TroyKomodo: The type is not public it makes it easier for the struct to use Signal inside a vec

lennartkloock: Oh it's not public, I missed that

lennartkloock:
?brawl merge

TroyKomodo:
?brawl retry

PR #396, opened 2025-03-03
rtmp/flv: E-RTMP support and refactor

lennartkloock:
- Implement E-RTMP support.
- Refactor scuffle-rtmp and scuffle-flv to be more ergonomic and efficient.

CLOUD-32
CLOUD-66


crates/rtmp/src/command_messages/netstream/reader.rs
```diff
@@ -0,0 +1,112 @@
...
+
+impl<'a> NetStreamCommand<'a> {
+    pub fn read(command_name: &str, decoder: &mut Amf0Decoder<'a>) -> Result<Option<Self>, CommandError> {
+        match command_name {
+            "play" => Ok(Some(Self::Play)),
+            "play2" => Ok(Some(Self::Play2)),
+            "deleteStream" => {
+                // skip command object
+                let Amf0Value::Null = decoder.decode_with_type(Amf0Marker::Null)? else {
+                    unreachable!();
```
TroyKomodo, line 18: why is this unreachable?

crates/rtmp/src/command_messages/netstream/reader.rs
```diff
@@ -0,0 +1,112 @@
...
+                };
+
+                let Amf0Value::Number(stream_id) = decoder.decode_with_type(Amf0Marker::Number)? else {
+                    unreachable!();
```
TroyKomodo, line 22: Same with this?

changes.d/pr-268.toml
```diff
@@ -0,0 +1,5 @@
+[[scuffle-rtmp]]
```
TroyKomodo, line 1: We should likely update this now

crates/rtmp/src/chunk/decoder.rs
```diff
@@ -38,18 +40,21 @@ pub struct ChunkDecoder {
...
 impl ChunkDecoder {
-    /// Sometimes a client will request a chunk size change.
+    /// Call when a client requests a chunk size change.
+    ///
+    /// Returns `false` if the chunk size is out of bounds.
+    /// The connection should be closed in this case.
     pub fn update_max_chunk_size(&mut self, chunk_size: usize) -> bool {
         // We need to make sure that the chunk size is within the allowed range.
-        // Returning false here will close the connection.
+        // Returning false here should close the connection.
```
TroyKomodo, line 57: When does it not?

crates/rtmp/src/chunk/define.rs
```diff
@@ -1,23 +1,14 @@
...
-    /// ChannelId for sending audio
-    Audio = 4,
-    /// ChannelId for sending video
-    Video = 5,
-}
+// common chunk stream ids
+
+pub const COMMAND_CHUNK_STREAM_ID: u32 = 3;
+pub const AUDIO_CHUNK_STREAM_ID: u32 = 4;
+pub const VIDEO_CHUNK_STREAM_ID: u32 = 5;
```
TroyKomodo, line 9: Should be a nutype enum for sure.

crates/rtmp/src/command_messages/netconnection/writer.rs
```diff
@@ -0,0 +1,118 @@
...
+                    ],
+                )?;
+            }
+            Self::CreateStreamResult { stream_id } => {
+                Amf0Encoder::encode_string(buf, "_result")?;
+                Amf0Encoder::encode_number(buf, transaction_id)?;
+                Amf0Encoder::encode_null(buf)?;
+                Amf0Encoder::encode_number(buf, stream_id)?;
+            }
+            _ => unimplemented!("the rtmp client is not implemented yet"),
```
TroyKomodo, line 44: Should likely be a command error not a panic, perhaps also remove the catch all and rather catch specific cases

crates/rtmp/src/command_messages/netconnection/reader.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use super::NetConnectionCommand;
+use crate::command_messages::errors::CommandError;
+
+impl<'a> NetConnectionCommand<'a> {
+    pub fn read(command_name: &str, decoder: &mut Amf0Decoder<'a>) -> Result<Option<Self>, CommandError> {
+        match command_name {
+            "connect" => {
+                let Amf0Value::Object(command_object) = decoder.decode_with_type(scuffle_amf0::Amf0Marker::Object)? else {
+                    unreachable!();
```
TroyKomodo, line 11: Why is this unreachable?

crates/rtmp/src/handshake/complex/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+pub mod digest;
+
+/// Complex Handshake Server
+/// Unfortunately there doesn't seem to be a good spec sheet for this.
+/// https://blog.csdn.net/win_lin/article/details/13006803 is the best I could find.
+pub struct ComplexHandshakeServer {
+    version: RtmpVersion,
+    requested_version: RtmpVersion,
+
+    pub(super) state: ServerHandshakeState,
```
TroyKomodo, line 23: does this need to be public at all?

crates/rtmp/src/handshake/mod.rs
```diff
@@ -1,12 +1,218 @@
...
-pub use self::define::{RTMP_HANDSHAKE_SIZE, ServerHandshakeState};
-pub use self::errors::*;
-pub use self::server::HandshakeServer;
+pub fn current_time() -> u32 {
+    let duration = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH);
+    match duration {
+        Ok(result) => result.as_nanos() as u32,
+        _ => 0,
+    }
+}
```
TroyKomodo, line 81: Use this or chrono?

crates/rtmp/src/messages/define.rs
```diff
@@ -1,38 +1,32 @@
...
+    Amf0Data { data: Bytes },
+    AudioData { data: Bytes },
+    VideoData { data: Bytes },
 }
 
-#[derive(Debug, PartialEq, Eq, Clone, Copy, FromPrimitive)]
+#[derive(Debug, PartialEq, Eq, Clone, Copy, num_derive::FromPrimitive)]
 #[repr(u8)]
-pub enum MessageTypeID {
+pub enum MessageType {
```
TroyKomodo, line 21: Should this be a nutype?

crates/rtmp/src/messages/parser.rs
```diff
@@ -1,53 +1,137 @@
...
-use crate::protocol_control_messages::ProtocolControlMessageReader;
-
-pub struct MessageParser;
+use crate::command_messages::Command;
+use crate::protocol_control_messages::ProtocolControlMessageSetChunkSize;
 
-impl MessageParser {
-    pub fn parse(chunk: &Chunk) -> Result<Option<RtmpMessageData<'_>>, MessageError> {
+impl MessageData<'_> {
+    pub fn parse(chunk: &Chunk) -> Result<Option<MessageData>, MessageError> {
```
TroyKomodo, line 8: returns Self?

crates/rtmp/src/messages/parser.rs
```diff
@@ -1,53 +1,137 @@
...
-                let chunk_size = ProtocolControlMessageReader::read_set_chunk_size(&chunk.payload)?;
-
-                Ok(Some(RtmpMessageData::SetChunkSize { chunk_size }))
-            }
-            // Metadata
-            MessageTypeID::DataAMF0 | MessageTypeID::DataAMF3 => Ok(Some(RtmpMessageData::AmfData {
+            MessageType::Video => Ok(Some(MessageData::VideoData {
                 data: chunk.payload.clone(),
             })),
             _ => Ok(None),
```
TroyKomodo, line 27: Perhaps an unknown variant?

crates/rtmp/src/user_control_messages/define.rs
```diff
@@ -1 +1,15 @@
-pub const RTMP_EVENT_STREAM_BEGIN: u16 = 0;
+#[derive(Debug, PartialEq, Eq, Clone, Copy, num_derive::FromPrimitive)]
+#[repr(u16)]
+pub enum EventType {
```
TroyKomodo, line 3: Should this be a nutype?

TroyKomodo:
I noticed in a lot of places we use `unreachable!` we should try avoid this unless we have a strong assertion that its truly unreachable.

crates/rtmp/src/command_messages/netstream/reader.rs
lennartkloock: Because `decode_with_type` returns an error when the data doesn't match the given marker. I guess I could remove the whole let-else though.

crates/rtmp/src/chunk/decoder.rs
lennartkloock: I changed it because this part of the code doesn't actually control if the connection will be closed. The caller decides what to do when false is returned.

crates/rtmp/src/command_messages/netstream/reader.rs
TroyKomodo: but what if it does not match?

crates/rtmp/src/messages/define.rs
lennartkloock: If everyone complies with the spec there are no other values allowed here. But it seems to be common practice to ignore the spec and do whatever you want so idk.

crates/rtmp/src/user_control_messages/define.rs
lennartkloock: Not sure, same as above

crates/rtmp/src/handshake/mod.rs
lennartkloock: This is fine, right? We don't have to introduce another dependency when we can do it with std.

lennartkloock:
?brawl try

lennartkloock:
?brawl try

crates/amf0/src/define.rs
```diff
@@ -51,6 +53,9 @@ pub enum Amf0Marker {
     AVMPlusObject = 0x11,
 }
 
+/// AMF0 object type.
+pub type Amf0Object<'a> = Cow<'a, [(Cow<'a, str>, Amf0Value<'a>)]>;
```
TroyKomodo, line 57: We should changelog entry this

crates/flv/src/audio/body/enhanced.rs
```diff
@@ -0,0 +1,222 @@
...
+pub enum ExAudioTagBody {
+    NoMultitrack {
+        audio_four_cc: AudioFourCc,
+        packet: AudioPacket,
+    },
+    ManyTracks(Vec<AudioTrack>),
+}
+
+impl ExAudioTagBody {
+    pub fn demux(header: &ExAudioTagHeader, reader: &mut io::Cursor<Bytes>) -> io::Result<Self> {
```
TroyKomodo, line 125: this function is very complex, it will be hard to test for sure

crates/flv/src/audio/header/enhanced.rs
```diff
@@ -0,0 +1,159 @@
...
+        SeguenceEnd = 2,
+        MultichannelConfig = 4,
+        Multitrack = 5,
+        ModEx = 7,
+    }
+}
+
+nutype_enum! {
+    pub enum AudioPacketModExType(u8) {
+        TimestampOffsetNano = 0,
```
TroyKomodo, line 24: What other types are there?

crates/flv/src/file.rs
```diff
@@ -20,14 +22,14 @@ impl FlvFile {
...
-    pub fn demux(reader: &mut std::io::Cursor<Bytes>) -> std::io::Result<Self> {
+    pub fn demux(reader: &mut std::io::Cursor<Bytes>) -> Result<Self, Error> {
         let header = FlvHeader::demux(reader)?;
 
         let mut tags = Vec::new();
         while reader.has_remaining() {
             // We don't care about the previous tag size, its only really used for seeking
             // backwards.
-            reader.read_u32::<BigEndian>()?;
+            reader.seek_relative(4)?;
```
TroyKomodo, line 32: This might be less efficient than reading the bytes. Since we are using io::Cursor this doesnt matter but if this was a buf io it would cause the buffer to be flushed.

crates/flv/src/script.rs
```diff
@@ -1,35 +1,246 @@
...
+pub enum OnMetaDataAudioCodecId {
+    Legacy(SoundFormat),
+    Enhanced(AudioFourCc),
+}
+
+impl OnMetaDataAudioCodecId {
+    fn from_amf0(value: &Amf0Value<'_>) -> Result<Self, Error> {
+        let n = value.as_number()? as u32;
+
+        if n > u8::MAX as u32 {
```
TroyKomodo, line 23: Explain why we use u8::MAX

crates/flv/src/script.rs
```diff
@@ -1,35 +1,246 @@
...
 
-        let data = amf0_reader
-            .decode_all()
-            .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "Invalid script data"))?;
+        match name.as_ref() {
+            // We might also want to handle "@setDataFrame" the same way as onMetaData.
+            // I'm not sure right now if that is the intended behavior though.
+            "onMetaData" => {
+                let value = amf0_reader.decode()?;
+                println!("{:?}", value);
```
TroyKomodo, line 217: stray println

crates/flv/src/script.rs
```diff
@@ -1,35 +1,246 @@
...
-            data: data.into_iter().map(|v| v.to_owned()).collect(),
-        })
+                let Amf0Value::Object(data) = value else { unreachable!() };
+                let data = OnMetaData::try_from(data)?;
+
+                Ok(Self::OnMetaData(Box::new(data)))
+            }
+            "onXMPData" => {
+                let value = amf0_reader.decode()?;
+                println!("{:?}", value);
```
TroyKomodo, line 226: stray println

crates/flv/src/script.rs
```diff
@@ -1,35 +1,246 @@
...
-        Ok(Self {
-            name: name.into_owned(),
-            data: data.into_iter().map(|v| v.to_owned()).collect(),
-        })
+                let Amf0Value::Object(data) = value else { unreachable!() };
+                let data = OnMetaData::try_from(data)?;
+
+                Ok(Self::OnMetaData(Box::new(data)))
+            }
+            "onXMPData" => {
```
TroyKomodo, line 224: What is XMPData

crates/flv/src/script.rs
```diff
@@ -1,35 +1,246 @@
...
+                let value = amf0_reader.decode()?;
+                println!("{:?}", value);
+
+                let Amf0Value::Object(data) = value else { unreachable!() };
+                let data = OnXmpData::try_from(data)?;
+
+                Ok(Self::OnXmpData(data))
+            }
+            _ => {
+                tracing::warn!(name = %name, "unknown script data");
```
TroyKomodo, line 234: i dont know how i feel about tracing here, since we return an other variant the caller can just add this if they want

crates/flv/src/script.rs
```diff
@@ -40,17 +251,65 @@ mod tests {
...
+            width[2],
+            width[3],
+            width[4],
+            width[5],
+            width[6],
+            width[7],
+            0,
+            0,
+            Amf0Marker::ObjectEnd as u8,
+        ];
```
TroyKomodo, line 289: Can you make this not reformat `#[rustfmt::skip]`
and then manually format it to make sense

crates/flv/src/script.rs
```diff
@@ -40,17 +251,65 @@ mod tests {
...
+        assert_eq!(metadata.can_seek_to_end, None);
+        assert_eq!(metadata.creationdate, None);
+        assert_eq!(metadata.duration, None);
+        assert_eq!(metadata.filesize, None);
+        assert_eq!(metadata.framerate, None);
+        assert_eq!(metadata.height, None);
+        assert_eq!(metadata.stereo, None);
+        assert_eq!(metadata.videocodecid, None);
+        assert_eq!(metadata.videodatarate, None);
+        assert_eq!(metadata.width, Some(1280.0));
```
TroyKomodo, line 313: Why not assert the whole struct at this point

crates/flv/src/tag.rs
```diff
@@ -124,15 +136,19 @@ impl FlvTagData {
...
             FlvTagType::Audio => Ok(FlvTagData::Audio(AudioData::demux(reader)?)),
-            FlvTagType::Video => Ok(FlvTagData::Video(VideoTagHeader::demux(reader)?)),
+            FlvTagType::Video => Ok(FlvTagData::Video(VideoData::demux(reader)?)),
             FlvTagType::ScriptData => Ok(FlvTagData::ScriptData(ScriptData::demux(reader)?)),
-            _ => Ok(FlvTagData::Unknown {
-                tag_type,
-                data: reader.extract_remaining(),
-            }),
+            _ => {
+                tracing::warn!(tag_type = ?tag_type, "unknown tag type");
```
TroyKomodo, line 145: see above about tracing comment

crates/flv/src/video/header/enhanced.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExVideoTagHeader {
+    pub video_packet_type: VideoPacketType,
+    pub video_packet_mod_exs: Vec<VideoPacketModEx>,
+    pub content: ExVideoTagHeaderContent,
+}
+
+impl ExVideoTagHeader {
+    #[allow(clippy::unusual_byte_groupings)]
```
TroyKomodo, line 101: what does this do?

crates/rtmp/src/command_messages/netstream/reader.rs
```diff
@@ -0,0 +1,237 @@
...
+                };
+
+                Ok(Some(Self::OnStatus {
+                    tc_url,
+                    level,
+                    code,
+                    description,
+                }))
+            }
+            _ => Ok(None),
```
TroyKomodo, line 102: maybe we should consider supporting the other commands ffmpeg sends and just returning to the user. instead of warning

crates/rtmp/src/command_messages/netstream/reader.rs
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+}
+
+impl From<&str> for NetStreamCommandPublishPublishingType {
```
TroyKomodo, line 107: Should this be `FromStr`

crates/rtmp/src/command_messages/reader.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        if let Some(command) = NetStreamCommand::read(&command_name, decoder)? {
+            return Ok(Self::NetStream(command));
+        }
+
+        Ok(Self::Unknown { command_name })
+    }
+}
+
+impl From<&str> for CommandResultLevel {
```
TroyKomodo, line 45: likewise

crates/rtmp/src/handshake/complex/digest.rs
```diff
@@ -30,7 +34,11 @@ impl<'a> DigestProcessor<'a> {
         }
     }
 
-    pub fn generate_and_fill_digest(&self, version: SchemaVersion) -> Result<(Bytes, [u8; 32], Bytes), DigestError> {
+    /// Generate and fill digest based on the schema version.
+    pub fn generate_and_fill_digest(
+        &self,
+        version: SchemaVersion,
+    ) -> Result<(Bytes, [u8; 32], Bytes), ComplexHandshakeError> {
```
TroyKomodo, line 41: Maybe we can look to return a struct that explains what each type is

crates/rtmp/src/lib.rs
```diff
@@ -1,16 +1,371 @@
...
+//! Unit tests are not yet fully implemented. Use at your own risk.
+//!
+//! ## License
+//!
+//! This project is licensed under the [MIT](./LICENSE.MIT) or [Apache-2.0](./LICENSE.Apache-2.0) license.
+//! You can choose between one of them if you use this work.
+//!
+//! `SPDX-License-Identifier: MIT OR Apache-2.0`
+#![cfg_attr(all(coverage_nightly, test), feature(coverage_attribute))]
+// #![deny(missing_docs)]
```
TroyKomodo, line 16: We should reenable this when its time to merge

crates/rtmp/src/session/server_session.rs
```diff
@@ -193,206 +200,188 @@ impl<S: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin> Session<S> {
...
                 .with_timeout(Duration::from_millis(2500))
-                .await??;
+                .await
+                .map_err(SessionError::Timeout)??;
 
             if n == 0 {
                 return Ok(false);
             }
+
+            let n = n.try_into().unwrap_or(u32::MAX);
```
TroyKomodo, line 210: this is ALWAYS less then u32::MAX since this is the number of bytes read and the read_buf is bound.

crates/rtmp/src/session/server_session.rs
```diff
@@ -193,206 +200,188 @@ impl<S: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin> Session<S> {
...
+
+            if (self.sequence_number % self.acknowledgement_window_size) + n >= self.acknowledgement_window_size {
```
TroyKomodo, line 212: What exactly does this logic mean

TroyKomodo (commented):
looks good so far, i think adding the tests and docs will be painful.

crates/flv/src/audio/header/enhanced.rs
lennartkloock: There are no other options defined in the spec. See enhanced-rtmp-v2, page 21.

crates/flv/src/script.rs
lennartkloock: It's from the legacy FLV spec. Annex E, section 6.

crates/flv/src/video/header/enhanced.rs
lennartkloock: It allows us to write things like this `0b0_111_0000` where the bits aren't grouped in blocks of 4. I group the bits like this to make it clear that they belong to different fields.

https://rust-lang.github.io/rust-clippy/master/index.html#unusual_byte_groupings

crates/rtmp/src/command_messages/netstream/reader.rs
lennartkloock: It cannot fail. `FromStr` returns a `Result`.

crates/rtmp/src/session/server_session.rs
lennartkloock: I'll add an explanation in the code

crates/rtmp/src/command_messages/netstream/reader.rs
TroyKomodo: then use fromStr with infallible

crates/rtmp/src/chunk/define.rs
```diff
@@ -1,23 +1,17 @@
...
-pub enum DefinedChunkStreamID {
-    /// ChannelId for sending commands
-    Command = 3,
-    /// ChannelId for sending audio
-    Audio = 4,
-    /// ChannelId for sending video
-    Video = 5,
+nutype_enum::nutype_enum! {
+    /// Common chunk stream ids
+    pub enum ChunkStreamId(u32) {
```
TroyKomodo, line 7: Should this be a nutype enum? this seems to just be for us used internall?

TroyKomodo: Also missing variant doc comments

crates/rtmp/src/chunk/error.rs
```diff
@@ -0,0 +1,13 @@
+#[derive(Debug, thiserror::Error)]
+pub enum ChunkReadError {
```
TroyKomodo, line 2: Missing docs

crates/rtmp/src/command_messages/mod.rs
```diff
@@ -0,0 +1,10 @@
+mod define;
+mod error;
+pub mod netconnection;
+pub mod netstream;
+pub mod on_status;
+mod reader;
+mod writer;
+
+pub use define::{CommandType, *};
```
TroyKomodo, line 9: why even export specifically the `CommndType`

crates/rtmp/src/command_messages/netstream/reader.rs
```diff
@@ -0,0 +1,160 @@
...
+                // skip command object
+                decoder.decode_with_type(Amf0Marker::Null)?;
+
+                let Amf0Value::String(publishing_name) = decoder.decode_with_type(Amf0Marker::String)? else {
+                    unreachable!();
+                };
+                let Amf0Value::String(publishing_type) = decoder.decode_with_type(Amf0Marker::String)? else {
+                    unreachable!();
+                };
+                let publishing_type = NetStreamCommandPublishPublishingType::from_str(&publishing_type).unwrap();
```
TroyKomodo, line 38: For now we can just do .expect("error is infalliable")

when https://doc.rust-lang.org/std/result/enum.Result.html#method.into_ok is stable we can switch.

crates/rtmp/src/command_messages/on_status/define.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct OnStatus<'a> {
+    pub code: Cow<'a, str>,
+    pub description: Option<Cow<'a, str>>,
+    pub level: CommandResultLevel,
+    pub others: Option<Amf0Object<'a>>,
+}
+
+// We can't use a nutype enum here because it would have to wrap a Cow<'a, str>.
```
TroyKomodo, line 15: Make a ticket for this maybe i can fix this

crates/rtmp/src/error.rs
```diff
@@ -0,0 +1,59 @@
+use crate::SessionError;
+use crate::chunk::ChunkReadError;
+use crate::command_messages::CommandError;
+use crate::handshake::ComplexHandshakeError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
TroyKomodo, line 7: Its generally bad practice to call the error `Error` because it will likely collide with other error types or the `std::error::Error` trait directly. rather use `RtmpError`

crates/rtmp/src/lib.rs
```diff
@@ -1,16 +1,380 @@
...
 mod messages;
-mod netconnection;
-mod netstream;
 mod protocol_control_messages;
 mod session;
 mod user_control_messages;
 
-pub use channels::{ChannelData, DataConsumer, DataProducer, PublishConsumer, PublishProducer, PublishRequest, UniqueID};
-pub use session::{Session, SessionError};
+pub use session::{Session, SessionData, SessionError, SessionHandler};
```
TroyKomodo, line 37: We have some mistakes here, we export `SessionError` but this doesnt rexport the sub error type.s

crates/rtmp/src/session/server_session.rs
```diff
@@ -179,10 +202,31 @@ impl<S: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin> Session<S> {
...
+    ///
+    /// Finish the handshake first by repeatedly calling [`drive_handshake`](Session::drive_handshake)
+    /// until it returns true before calling this function.
+    ///
+    /// Returns true if the session is still active, false if the client has closed the connection.
+    async fn drive(&mut self) -> Result<bool, crate::error::Error> {
+        // Send a reconnect request if we haven't yet, the client supports it and the context is cancelled
+        if !self.reconnect_request_sent
+            && self.caps_ex.is_some_and(|c| c.intersects(CapsExMask::Reconnect))
+            && self.ctx.as_ref().is_some_and(|ctx| ctx.is_done())
```
TroyKomodo, line 217: I dont like this. I think this will be hard to use. Ideally we should have a function that we can call to send a reconnect.

TroyKomodo (changes requested):

crates/flv/Cargo.toml
```diff
@@ -19,6 +19,9 @@ bytes = "1.5"
 num-traits = "0.2"
 num-derive = "0.4"
 thiserror = "2.0"
+bitmask-enum = "2.2.5"
+tracing = "0.1"
```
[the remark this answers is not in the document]
lennartkloock: bitmask-enum or tracing?

crates/rtmp/src/protocol_control_messages/define.rs
```diff
@@ -0,0 +1,37 @@
...
+pub struct ProtocolControlMessageWindowAcknowledgementSize {
+    pub acknowledgement_window_size: u32,
+}
+
+/// 5.4.5. Set Peer Bandwidth (6)
+#[derive(Debug)]
+pub struct ProtocolControlMessageSetPeerBandwidth {
+    pub acknowledgement_window_size: u32,
+    pub limit_type: ProtocolControlMessageSetPeerBandwidthLimitType,
+}
```
philipch07: Perhaps a brief description for the two fields?

crates/rtmp/src/chunk/mod.rs
```diff
@@ -1,12 +1,108 @@
...
+pub enum ChunkType {
+    /// Chunk type 0 - 5.3.1.2.1
+    Type0 = 0,
+    /// Chunk type 1 - 5.3.1.2.2
+    Type1 = 1,
+    /// Chunk type 2 - 5.3.1.2.3
+    Type2 = 2,
+    /// Chunk type 3 - 5.3.1.1.4
+    Type3 = 3,
+}
```
philipch07, line 30: Perhaps a brief description of what the types do or what makes them different?

crates/rtmp/src/command_messages/netconnection/reader.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use super::{CapsExMask, NetConnectionCommand, NetConnectionCommandConnect};
+use crate::command_messages::error::CommandError;
+
+impl<'a> NetConnectionCommand<'a> {
+    pub fn read(command_name: &str, decoder: &mut Amf0Decoder<'a>) -> Result<Option<Self>, CommandError> {
+        match command_name {
+            "connect" => {
+                let Amf0Value::Object(command_object) = decoder.decode_with_type(scuffle_amf0::Amf0Marker::Object)? else {
+                    unreachable!();
```
philipch07, line 11: I think it could be useful to leave a comment explaining why the line is truly unreachable + make a jira ticket with all unreachable instances and saying that we should improve amf0 to avoid these.

crates/rtmp/src/session/server_session.rs
```diff
@@ -30,115 +49,117 @@ pub struct Session<S> {
...
     /// connection (using different stream keys) and or play multiple streams
     /// per RTMP connection (using different stream keys) as per the RTMP spec.
-    app_name: Option<String>,
-
-    /// This is a unique id for this session
-    /// This is issued when the client connects to the server
-    uid: Option<UniqueID>,
-
+    app_name: Option<Box<str>>,
+    caps_ex: Option<CapsExMask>,
```
philipch07, line 53: Missing docs for `caps_ex`?

crates/rtmp/src/user_control_messages/mod.rs
```diff
@@ -1,9 +1,17 @@
...
+    pub enum EventType(u16) {
+        StreamBegin = 0,
+        StreamEOF = 1,
+        StreamDry = 2,
+        SetBufferLength = 3,
+        StreamIsRecorded = 4,
+        PingRequest = 6,
+        PingResponse = 7,
+    }
+}
```
philipch07, line 13: Perhaps add a comment explaining where this enum comes from in the docs if there's a section that this is from.

crates/transmuxer/src/errors.rs
```diff
@@ -1,44 +1,27 @@
...
-            }
-            Self::NoSequenceHeaders => write!(f, "no sequence headers"),
-            Self::IO(err) => write!(f, "io error: {}", err),
-        }
-    }
+    #[error("io error: {0}")]
+    Io(#[from] io::Error),
+    #[error("flv error: {0}")]
+    Flv(#[from] scuffle_flv::error::FlvError),
 }
```
philipch07, line 27: This implementation seems pretty nice. Should we add a jira ticket to use this style of error declaration in our other crates?

crates/flv/src/audio/body/enhanced.rs
```diff
@@ -0,0 +1,755 @@
...
+nutype_enum! {
+    /// Channel mappings enum
+    ///
+    /// See <https://en.wikipedia.org/wiki/Surround_sound#Standard_speaker_channels> and
+    /// <https://en.wikipedia.org/wiki/22.2_surround_sound> for more information.
+    pub enum AudioChannel(u8) {
+        // commonly used speaker configurations
+        //
+        // see - <https://en.wikipedia.org/wiki/Surround_sound#Standard_speaker_channels>
+
```
philipch07, line 38: I don't think we need this comment since it's explained above already (or maybe you forgot to delete this).

crates/flv/src/audio/body/enhanced.rs
```diff
@@ -0,0 +1,755 @@
...
+}
+
+/// Mask used to indicate which channels are present in the stream.
+///
+/// See <https://en.wikipedia.org/wiki/Surround_sound#Standard_speaker_channels> and
+/// <https://en.wikipedia.org/wiki/22.2_surround_sound> for more information.
+#[bitmask_enum::bitmask(u32)]
+pub enum AudioChannelMask {
+    // masks for commonly used speaker configurations
+    // <https://en.wikipedia.org/wiki/Surround_sound#Standard_speaker_channels>
```
philipch07, line 106: same as above

crates/flv/src/audio/body/enhanced.rs
```diff
@@ -0,0 +1,755 @@
...
+    BottomFrontRight = 0x800000,
+}
+
+/// Multichannel configuration
+///
+/// Describes the configuration of the audio channels in a multichannel audio stream.
+///
+/// Contained in an [`AudioPacket::MultichannelConfig`].
+#[derive(Debug, Clone, PartialEq)]
+pub enum MultichannelConfigOrder {
```
philipch07, line 167: Can you add a reference to where this is described in the spec (if applicable?)

crates/flv/src/script.rs
```diff
@@ -1,56 +1,540 @@
...
+                    for (key, value) in object.iter() {
+                        map.insert(key.to_string(), value.to_owned());
+                    }
+
+                    video_track_id_info_map = Some(map);
+                }
+                _ => {
+                    other.insert(key.to_string(), value.to_owned());
+                }
+            }
```
philipch07, line 216: Yeah we should probably impl serde support. What is necessary for that to happen? Also perhaps a jira ticket for this since it sounds like it's going to be a large task since you didn't add it yet.

crates/flv/src/script.rs
```diff
@@ -1,56 +1,540 @@
...
+/// Defined by:
+/// - Legacy FLV spec, Annex E.6
+#[derive(Debug, Clone, PartialEq)]
+pub struct OnXmpData {
+    /// XMP metadata, formatted according to the XMP metadata specification.
+    ///
+    /// For further details, see [www.adobe.com/devnet/xmp/pdfs/XMPSpecificationPart3.pdf](https://web.archive.org/web/20090306165322/https://www.adobe.com/devnet/xmp/pdfs/XMPSpecificationPart3.pdf).
+    live_xml: Option<String>,
+    /// Any other metadata contained in the script data.
+    other: HashMap<String, Amf0Value<'static>>,
```
philipch07, line 253: A couple examples of what metadata counts as `other` would be nice here.

crates/rtmp/src/chunk/reader.rs
```diff
@@ -0,0 +1,1080 @@
...
+    }
+
+    /// This function is used to read a chunk from the buffer.
+    ///
+    /// Returns:
+    /// - `Ok(None)` if the buffer does not contain enough data to read a full chunk.
+    /// - `Ok(Some(Chunk))` if a full chunk is read.
+    /// - `Err(ChunkReadError)` if there is an error decoding a chunk. The connection should be closed.
+    ///
+    /// # See also
```
philipch07, line 73: Should this be `##` instead of `#`?

crates/rtmp/src/chunk/reader.rs
```diff
@@ -0,0 +1,1080 @@
...
+                    .ok_or(crate::error::RtmpError::ChunkRead(
+                        ChunkReadError::MissingPreviousChunkHeader(header.chunk_stream_id),
+                    ))?
+                    .clone();
+
+                // Now this is truely stupid.
+                // If the PREVIOUS HEADER is extended then we now waste an additional 4 bytes to
+                // read the timestamp. Why not just read the timestamp in the previous header if
+                // it is extended? I guess the spec writers had some reason and its obviously
+                // way above my knowledge.
```
philipch07, line 493: image

crates/rtmp/src/command_messages/reader.rs
```diff
@@ -0,0 +1,75 @@
...
+        if let Some(command) = NetConnectionCommand::read(&command_name, decoder)? {
+            return Ok(Self::NetConnection(command));
+        }
+
+        if let Some(command) = NetStreamCommand::read(&command_name, decoder)? {
+            return Ok(Self::NetStream(command));
+        }
+
+        Ok(Self::Unknown { command_name })
+    }
```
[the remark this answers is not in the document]
lennartkloock: What exactly do you think is wrong with this?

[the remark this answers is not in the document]
philipch07: I'm wondering if there's a better way to handle:
```rust
if let Some(command) = fn1::do_something() {
    return Ok(Self::fn1(command));
}

if let Some(command) = fn2::do_something() {
    return Ok(Self::fn2(command));
}

Ok(Self::fn3 { command_name })
```
Since it seems kinda similar to: `if cond1 -> fn1`, `elif cond2 -> fn2`, `else fn3` if that makes sense. It's not necessary that the current structure needs to change, but I was just wondering if there's another way to write out that logic.

crates/rtmp/src/handshake/complex/digest.rs
```diff
@@ -1,17 +1,47 @@
...
+/// Use [`DigestProcessor::generate_and_fill_digest`] to create a `DigestResult`
+/// and [`DigestResult::write_to`] to write the result to a buffer.
+pub struct DigestResult {
+    /// The left part.
+    pub left: Bytes,
+    /// The digest.
+    pub digest: [u8; 32],
+    /// The right part.
+    pub right: Bytes,
+}
```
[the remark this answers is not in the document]
lennartkloock: To be honest, I'm not really sure what it means because the only source for this algorithm is a Chinese forum post. ([Yes I'm not joking](https://blog.csdn.net/win_lin/article/details/13006803))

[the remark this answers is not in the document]
philipch07: @TroyKomodo would you happen to recall what section of the post this is from?

crates/rtmp/src/chunk/define.rs
lennartkloock: Why not? It seems like the right use case for a nutype enum to me here.

crates/transmuxer/src/errors.rs
lennartkloock: It's already used in a lot of our crates. scuffle-ffmpeg and scuffle-http for example.

philipch07: Hmm, I wonder if there are any instances of us writing errors the old way. I would have to check another time.

crates/flv/src/script.rs
lennartkloock: I'm not sure yet what is necessary for that to happen. The Jira ticket is CLOUD-91.

crates/flv/src/video/body/enhanced/metadata.rs
```diff
@@ -0,0 +1,386 @@
...
+
+use crate::error::Error;
+
+/// Color configuration metadata.
+///
+/// > `colorPrimaries`, `transferCharacteristics` and `matrixCoefficients` are defined
+/// > in ISO/IEC 23091-4/ITU-T H.273. The values are an index into
+/// > respective tables which are described in "Colour primaries",
+/// > "Transfer characteristics" and "Matrix coefficients" sections.
+/// > It is RECOMMENDED to provide these values.
```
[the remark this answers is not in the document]
lennartkloock: It's a markdown quote and I used it whenever I quoted paragraphs from the spec.

crates/flv/src/video/body/enhanced/metadata.rs
```diff
@@ -0,0 +1,386 @@
...
+                    ("greenY".into(), 0.0.into()),
+                    ("blueX".into(), 0.0.into()),
+                    ("blueY".into(), 0.0.into()),
+                    ("whitePointX".into(), 0.0.into()),
+                    ("whitePointY".into(), 0.0.into()),
+                    ("maxLuminance".into(), 0.0.into()),
+                    ("minLuminance".into(), 0.0.into()),
+                ])),
+            ),
+        ]);
```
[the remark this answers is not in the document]
lennartkloock: It's not really avoidable because we have to convert the values into `Cow`s and `Amf0Value`s. This is the shortest way of doing it.

crates/rtmp/src/chunk/define.rs
```diff
@@ -32,51 +26,54 @@ pub enum ChunkType {
...
+                         * write 0xFFFFFF) */
+    /// The length of the message.
+    pub msg_length: u32, // 3 bytes
+    /// The type of the message.
+    pub msg_type_id: MessageType, // 1 byte
+    /// The stream id of the message.
+    pub msg_stream_id: u32, // 4 bytes
+    /// Whether the timestamp is extended.
+    pub was_extended_timestamp: bool, // used for reading the header only
 }
```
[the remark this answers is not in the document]
lennartkloock: I don't think the bytes sizes are relevant here because the user usually doesn't care about how exactly this is implemented.

crates/rtmp/src/command_messages/reader.rs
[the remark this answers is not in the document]
lennartkloock: You could use nested if-let or match statements but that would get messy very quickly.

crates/rtmp/src/handshake/simple.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+            ServerHandshakeState::Finish => {}
+        }
+
+        Ok(())
+    }
+
+    fn read_c0(&mut self, input: &mut io::Cursor<Bytes>) -> Result<(), crate::error::Error> {
+        // Version (8 bits): In C0, this field identifies the RTMP version
+        //  requested by the client.
```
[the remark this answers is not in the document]
lennartkloock: No, this is not normal. I will fix it.

crates/flv/src/script.rs
lennartkloock: This is a catch-all map which means all other parsed values get added to this map. There aren't really any examples because we don't know what the client will send.

crates/rtmp/src/chunk/reader.rs
lennartkloock: I think it's fine like this. I looked at the std library docs and they also use first level headings everywhere.

crates/flv/src/script.rs
lennartkloock: This is a catch-all map which means all other parsed values get added to this map. There aren't really any examples because we don't know what the client will send.

crates/transmuxer/src/errors.rs
lennartkloock: I just checked and I couldn't find any. This was probably the only case.

philipch07: Oh nice, thanks for checking!

crates/rtmp/src/session/server_session.rs
lennartkloock: The current design takes ownership of the session when calling the `run` function. I'm not sure how to redesign it so that the caller can keep a reference to it.

crates/rtmp/src/chunk/define.rs
TroyKomodo: well we are the only ones who use this type its not from the wire

crates/rtmp/src/handshake/complex/digest.rs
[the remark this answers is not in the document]
TroyKomodo: i know I said initially we should use a struct. however this is exactly what `bytes::Buf` does.

We could replace this with 

```
left.chain(digest).chain(right)
```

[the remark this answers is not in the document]
philipch07: Spoke offline (for reference the section is "complex handshake C1S1 algorithm" (complex handshake C1S1算法) which basically says: the digest generation is comprised of 3 parts that sum to exactly 1536 bytes. The first part is `n` bytes, the second part is the 32 byte digest, and the third part is `m` bytes such that `n + 32 + m = 1536`.

Then the article explains how they do some other stuff, but I guess this can suffice for more context in the doc comment for now... I'm a little surprised there's so little publicly accessible information about this, but oh well.

crates/ffmpeg/src/filter_graph.rs
```diff
@@ -469,7 +469,7 @@ mod tests {
             .expect("Failed to set input for the filter graph");
 
         assert!(
-            input_parser.graph.as_ptr() == filter_graph.as_ptr(),
+            std::ptr::eq(input_parser.graph.as_ptr(), filter_graph.as_ptr()),
```
TroyKomodo, line 472: this change shouldnt be here, rebase this pr

crates/flv/src/video/body/enhanced/mod.rs
```diff
@@ -0,0 +1,600 @@
...
+                let data = reader.extract_bytes(size_of_video_track.unwrap_or(reader.remaining()))?;
+
+                let seq_start = match video_four_cc {
+                    VideoFourCc::Av1 => {
+                        let record = AV1CodecConfigurationRecord::demux(&mut io::Cursor::new(data))?;
+                        VideoPacketSequenceStart::Av1(record)
+                    }
+                    VideoFourCc::Avc => {
+                        let record = AVCDecoderConfigurationRecord::parse(&mut io::Cursor::new(data))?;
+                        VideoPacketSequenceStart::Avc(record)
```
TroyKomodo, line 138: Is this not just the legacy version? does anyone use this with enhanced?

crates/flv/src/video/body/enhanced/mod.rs
```diff
@@ -0,0 +1,600 @@
...
+                Ok(Self::SequenceStart(seq_start))
+            }
+            VideoPacketType::Mpeg2TsSequenceStart => {
+                let data = reader.extract_bytes(size_of_video_track.unwrap_or(reader.remaining()))?;
+
+                let seq_start = match video_four_cc {
+                    VideoFourCc::Av1 => {
+                        let descriptor = AV1VideoDescriptor::demux(&mut io::Cursor::new(data))?;
+                        VideoPacketMpeg2TsSequenceStart::Av1(descriptor)
+                    }
```
TroyKomodo, line 156: Do we need tests for this?

crates/rtmp/examples/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "scuffle-rtmp-examples"
+version = "0.1.0"
+edition = "2024"
+description = "Examples for scuffle-rtmp"
+repository = "https://github.com/scufflecloud/scuffle"
+authors = ["Scuffle <opensource@scuffle.cloud>"]
+readme = "README.md"
+documentation = "https://docs.rs/scuffle-rtmp-examples"
+license = "MIT OR Apache-2.0"
+keywords = ["rtmp", "server", "streaming"]
```
TroyKomodo, line 11: remove all of this and replace it with `publish = false` or this crate will be published to `crates.io`

crates/rtmp/src/command_messages/netstream/reader.rs
```diff
@@ -0,0 +1,170 @@
...
+                    // TODO: CLOUD-91
+                    unreachable!();
+                };
+                // TODO: change expect to into_ok once stabliized
+                // https://doc.rust-lang.org/std/result/enum.Result.html#method.into_ok
+                let publishing_type =
+                    NetStreamCommandPublishPublishingType::from_str(&publishing_type).expect("infalible error");
+
+                if let NetStreamCommandPublishPublishingType::Unknown(publishing_type) = &publishing_type {
+                    tracing::warn!(publishing_type = ?publishing_type, "unknown publishing type in publish command");
```
TroyKomodo, line 51: bad tracing warn

crates/rtmp/src/command_messages/reader.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn read(command_name: Cow<'a, str>, decoder: &mut Amf0Decoder<'a>) -> Result<Self, CommandError> {
+        if let Some(command) = NetConnectionCommand::read(&command_name, decoder)? {
+            return Ok(Self::NetConnection(command));
+        }
+
+        if let Some(command) = NetStreamCommand::read(&command_name, decoder)? {
+            return Ok(Self::NetStream(command));
+        }
+
+        Ok(Self::Unknown { command_name })
```
TroyKomodo, line 48: when we return unknown we should decode all values from the amf0 decoder

crates/rtmp/src/command_messages/writer.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Converts the [`CommandResultLevel`] to a [`String`] by taking ownership.
+    pub fn into_string(self) -> String {
+        match self {
+            CommandResultLevel::Warning => "warning".to_string(),
+            CommandResultLevel::Status => "status".to_string(),
+            CommandResultLevel::Error => "error".to_string(),
+            CommandResultLevel::Unknown(s) => s,
+        }
+    }
+}
```
TroyKomodo, line 34: Is this not just `AsRef<str>` and `std::fmt::Display`

crates/rtmp/src/lib.rs
```diff
@@ -1,16 +1,379 @@
...
+//!
+//! ## Specifications
+//!
+//! | Name | Version | Link | Comments |
+//! | --- | --- | --- | --- |
+//! | Adobe’s Real Time Messaging Protocol | `1.0` | <https://github.com/veovera/enhanced-rtmp/blob/main/docs/legacy/rtmp-v1-0-spec.pdf> | Refered to as 'Legacy RTMP spec' in this documentation |
+//! | Enhancing RTMP, FLV | `v1-2024-02-29-r1` | <https://github.com/veovera/enhanced-rtmp/blob/main/docs/enhanced/enhanced-rtmp-v1.pdf> | |
+//! | Enhanced RTMP | `v2-2024-10-22-b1` | <https://github.com/veovera/enhanced-rtmp/blob/main/docs/enhanced/enhanced-rtmp-v2.pdf> | Refered to as 'Enhanced RTMP spec' in this documentation |
+//!
+//! ## Status
```
TroyKomodo, line 11: maybe add a doc example

crates/rtmp/src/messages/reader.rs
```diff
@@ -0,0 +1,154 @@
...
+        match chunk.message_header.msg_type_id {
+            // Protocol Control Messages
+            MessageType::SetChunkSize => {
+                let data = ProtocolControlMessageSetChunkSize::read(&chunk.payload)?;
+                Ok(Self::SetChunkSize(data))
+            }
+            MessageType::WindowAcknowledgementSize => {
+                let data = ProtocolControlMessageWindowAcknowledgementSize::read(&chunk.payload)?;
+                Ok(Self::SetAcknowledgementWindowSize(data))
+            }
```
TroyKomodo, line 22: Is this never called?

crates/rtmp/src/session/server/handler.rs
```diff
@@ -0,0 +1,52 @@
...
+    Amf0 {
+        /// Timestamp of the data.
+        timestamp: u32,
+        /// Data.
+        data: Bytes,
+    },
+}
+
+/// Handler for session events.
+pub trait SessionHandler {
```
TroyKomodo, line 34: we should have another function for unhandled commands.

crates/rtmp/src/session/server/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+            MessageData::VideoData { data } => {
+                self.handler
+                    .on_data(stream_id, SessionData::Video { timestamp, data })
+                    .await?;
+            }
+            MessageData::DataAmf0 { data } => {
+                self.handler.on_data(stream_id, SessionData::Amf0 { timestamp, data }).await?;
+            }
+            MessageData::Other { msg_type_id, data } => {
+                tracing::warn!(msg_type_id = ?msg_type_id, data = ?data, "unknown message type");
```
TroyKomodo, line 331: we should report to the handler i think.

crates/rtmp/src/session/server/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+            }) => {
+                self.on_command_publish(stream_id, command.transaction_id, &publishing_name, publishing_type)
+                    .await?;
+            }
+            CommandType::NetStream(NetStreamCommand::CloseStream) => {
+                // Not sure what this is for
+            }
+            CommandType::Unknown { command_name } => {
+                tracing::warn!(command_name = ?command_name, "unknown command");
+            }
```
TroyKomodo, line 380: we should report to the handler

TroyKomodo (changes requested):

crates/flv/src/video/body/enhanced/mod.rs
lennartkloock: H.264 is support in the enhanced spec too. I don't think anyone is using it but it is supported.

crates/rtmp/src/messages/reader.rs
lennartkloock: Not sure if this is actually ever sent by ffmpeg or obs. I added a test for the coverage.

crates/flv/src/video/body/enhanced/mod.rs
lennartkloock: Not sure, I was too lazy to manually construct a valid Av1 video descriptor

crates/rtmp/src/session/server_session.rs
lennartkloock: CLOUD-95

TroyKomodo (approved):
Very nice!!

TroyKomodo:
?brawl merge

PR #397, opened 2025-03-03
add base svelte template to monorepo

lucassshanks:
CLOUD-78 

add default svelte repo

using this as root branch for f/e dashboard so keeping it up to date

Closing this in favour of larger initial svelte PR


lennartkloock:
why is the docs deploy failing here?

TroyKomodo:
> why is the docs deploy failing here?

i think its something to do with pnpm workspace breaking the pnpm install somehow same issue on #434

PR #398, opened 2025-03-04
fix clippy lints

TroyKomodo:
Seems that clippy got updated and discovered some new lints.

### DIFF
--- a/crates/http/src/backend/hyper/stream.rs
+++ b/crates/http/src/backend/hyper/stream.rs
@@ -7,7 +7,7 @@ pub enum Stream {
     Tcp(tokio::net::TcpStream),
     #[cfg(feature = "tls-rustls")]
     #[cfg_attr(docsrs, doc(cfg(feature = "tls-rustls")))]
-    Tls(tokio_rustls::server::TlsStream<tokio::net::TcpStream>),
+    Tls(Box<tokio_rustls::server::TlsStream<tokio::net::TcpStream>>),
 }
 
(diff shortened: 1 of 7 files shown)


crates/http/src/backend/hyper/stream.rs
```diff
@@ -7,7 +7,7 @@ pub enum Stream {
     Tcp(tokio::net::TcpStream),
     #[cfg(feature = "tls-rustls")]
     #[cfg_attr(docsrs, doc(cfg(feature = "tls-rustls")))]
-    Tls(tokio_rustls::server::TlsStream<tokio::net::TcpStream>),
+    Tls(Box<tokio_rustls::server::TlsStream<tokio::net::TcpStream>>),
```
TroyKomodo, line 10: not sure if this is breaking or not.

philipch07 (approved):

crates/http/src/backend/hyper/stream.rs
philipch07: I was going to ask about this; what does wrapping with a Box do?

TroyKomodo: Box'd values are pointers to heap memory

philipch07: Do u want to try running `just grind` locally to check if it has any impacts on memory then?

TroyKomodo:
?brawl merge

PR #407, opened 2025-03-13
Implement callbacks to configure the sockets used as listeners

DrSloth:
Hello, i have recently come across a situation where i wanted to use SO_REUSEPORT on the HttpServers. 
I added options to the HttpServerBuilder that can be set to callbacks which can configure `socket2::Socket` instances that will then be used as listeners.

### DIFF
--- a/crates/http/src/backend/h3/mod.rs
+++ b/crates/http/src/backend/h3/mod.rs
@@ -12,6 +12,7 @@ use utils::copy_response_body;
 
 use crate::error::Error;
 use crate::service::{HttpService, HttpServiceFactory};
+use crate::server::ConfigureSocketCallback;
 
 pub mod body;
 mod utils;
@@ -36,6 +37,8 @@ pub struct Http3Backend<F> {
     /// Use `[::]` for a dual-stack listener.
     /// For example, use `[::]:80` to bind to port 80 on both IPv4 and IPv6.
     bind: SocketAddr,
+    /// Callback to configure socket
+    configure_sock: Option<ConfigureSocketCallback>,
     /// rustls config.
     ///
     /// Use this field to set the server into TLS mode.
@@ -67,7 +70,27 @@ where
         let server_config = h3_quinn::quinn::ServerConfig::with_crypto(Arc::new(crypto));
 
         // Bind the UDP socket
-        let socket = std::net::UdpSocket::bind(self.bind)?;
+        let socket = {
+            let mut sock = socket2::Socket::new(
+                match self.bind {
+                    SocketAddr::V4(_) => socket2::Domain::IPV4,
+                    SocketAddr::V6(_) => socket2::Domain::IPV6,
+                },
+                socket2::Type::DGRAM,
+                Some(socket2::Protocol::UDP),
+            )?;
+
+            sock.set_nonblocking(true)?;
+            sock.set_only_v6(false)?;
+            if let Some(cfg_fn) = self.configure_sock.as_ref() {
+                sock = cfg_fn.call(sock)?;
+            }
+            
+            sock.bind(&socket2::SockAddr::from(self.bind))?;
+            sock.listen(128)?;
+
+            std::net::UdpSocket::from(sock)
+        };
 
         // Runtime for the quinn endpoint
         let runtime = h3_quinn::quinn::default_runtime().ok_or_else(|| io::Error::other("no async runtime found"))?;
--- a/crates/http/src/backend/hyper/mod.rs
+++ b/crates/http/src/backend/hyper/mod.rs
@@ -8,6 +8,7 @@ use tracing::Instrument;
 
 use crate::error::Error;
 use crate::service::{HttpService, HttpServiceFactory};
+use crate::server::ConfigureSocketCallback;
 
 mod handler;
 mod stream;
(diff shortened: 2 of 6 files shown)


crates/http/src/backend/hyper/mod.rs
```diff
@@ -79,7 +82,27 @@ where
         }
 
         // We have to create an std listener first because the tokio listener isn't clonable
-        let listener = tokio::net::TcpListener::bind(self.bind).await?.into_std()?;
+        let listener = {
```
TroyKomodo, line 85: I suspect this might not work on windows. Previously when adding windows support we found that if the listener was constructed outside of tokio it would block the eventloop even if non-blocking was set to true. We didnt investigate this further than that, but perhaps this might be a good time to understand why this behaviour was the case when using  `std::net::TcpListener::bind`

TroyKomodo:
If you can format your code using `just fmt` I can run the windows tests to see if the socket2 bind with non-blocking actually works on windows environments.

crates/http/src/server.rs
```diff
@@ -40,6 +41,10 @@ pub struct HttpServer<F> {
     #[cfg(feature = "http3")]
     #[cfg_attr(docsrs, doc(cfg(feature = "http3")))]
     enable_http3: bool,
+    /// Callback to configure socket used for http1 and http2
+    configure_h12_sock: Option<ConfigureSocketCallback>,
+    /// Callback to configure socket used for http3
+    configure_h3_sock: Option<ConfigureSocketCallback>,
```
lennartkloock, line 47: Can we combine this into one field? At this level we try to abstract away the differences.

crates/http/src/backend/hyper/mod.rs
DrSloth: I have run just fmt now. I haven't had this problem on windows yet at least when using the socket2 crate but my time spent in windows is not a lot.

crates/http/src/server.rs
DrSloth: It would be possible. Most configuration done is TCP only but it may be a niche use case to set different buffer sizes for TCP and UDP.

crates/http/src/backend/h3/mod.rs
```diff
@@ -67,7 +70,27 @@ where
...
+                    SocketAddr::V4(_) => socket2::Domain::IPV4,
+                    SocketAddr::V6(_) => socket2::Domain::IPV6,
+                },
+                socket2::Type::DGRAM,
+                Some(socket2::Protocol::UDP),
+            )?;
+
+            sock.set_nonblocking(true)?;
+            sock.set_only_v6(false)?;
+            if let Some(cfg_fn) = self.configure_sock.as_ref() {
```
philipch07, line 85: It may be useful to comment that this sets the callback. A short comment will be helpful for new readers + glance value.

philipch07:
?brawl try

philipch07:
Oh I missed some files when reviewing since I'm currently on mobile.

When you get the chance, please run `just workspace-hack` so it will pass the hakari stage.

crates/http/src/backend/h3/mod.rs
lennartkloock: What do you mean? This doesn't set the callback.

philipch07: Ah I misunderstood the line (86) so I thought this set the callback.

lennartkloock:
?brawl try

TroyKomodo:
?brawl try

crates/http/src/server.rs
```diff
@@ -297,3 +311,27 @@ where
...
+
+    /// Create a new `ConfigureSocketCallback` from the given callback function.
+    pub fn call(&self, sock: socket2::Socket) -> std::io::Result<socket2::Socket> {
+        (self.0)(sock)
+    }
+}
+
+impl std::fmt::Debug for ConfigureSocketCallback {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "ConfigureSocketCallback ")
```
TroyKomodo, line 335: extra space here

crates/http/src/server.rs
```diff
@@ -40,6 +41,14 @@ pub struct HttpServer<F> {
     #[cfg(feature = "http3")]
     #[cfg_attr(docsrs, doc(cfg(feature = "http3")))]
     enable_http3: bool,
+    /// Callback to configure socket used for http1 and http2
+    #[cfg(any(feature = "http1", feature = "http2"))]
+    #[cfg_attr(docsrs, doc(cfg(any(feature = "http1", feature = "http2"))))]
+    configure_h12_sock: Option<ConfigureSocketCallback>,
```
TroyKomodo, line 47: can we make bon use an `Into` impl for this and then impl this for any `Fn<...>`

TroyKomodo:
@DrSloth did you still plan to do this pr?

DrSloth:
> @DrSloth did you still plan to do this pr?

Yes, i am very sorry. Some other projects came up that i need to attend right now.
Only unit tests are missing.
I am currently hunting for some error that prevents me from writing a unit test that expects failure. Binding twice to the same udp port for http3 fails with a aws-lc error.

PR #416, opened 2025-03-24
fix workspace-deps xtask tool

TroyKomodo:
The bug was related to how toml_edit replaces text, we should reuse the existing dep instead of creating a new one. This is likely a bug in toml_edit itself, but doing it this way avoids it on our end. :)


philipch07 (approved):

philipch07:
?brawl merge

PR #418, opened 2025-03-24
docs(http): improve examples

lennartkloock:
- Add echo examples to `scuffle-http` because it's the classic server example

CLOUD-96


philipch07 (approved):
Looks good but might want to double check w/ troy since im not too familiar with server stuff, unless you're confident that this is fine.

TroyKomodo:
?brawl merge

PR #420, opened 2025-03-24
refactor: remove example crates

lennartkloock:
- Remove example crates and include examples in the library crates themselves
- Fix some dependencies

CLOUD-97


TroyKomodo:
Do we need to specify each example in the examples directory? I know cargo automatically resolves tests in a test directory I assume it does the same for examples.

lennartkloock:
> Do we need to specify each example in the examples directory? I know cargo automatically resolves tests in a test directory I assume it does the same for examples.

You're right, we don't have to. But when we don't explicitly set the name of the example it defaults to the example's file name. So we either rename the example files or we keep it like this.

TroyKomodo (approved):

TroyKomodo:
?brawl merge

lennartkloock:
?brawl try

lennartkloock:
idk what I'm doing here, I couldve just ran it myself

crates/bootstrap/examples/cli.rs
```diff
@@ -45,7 +45,7 @@ struct Global {
     pub config: Config,
 }
 
-#[derive(serde_derive::Deserialize, Debug, smart_default::SmartDefault)]
```
TroyKomodo, line 48: We should make a ticket to remove the `derive` feature from serde. https://github.com/serde-rs/serde/issues/2584

TroyKomodo (approved):

PR #421, opened 2025-03-27
feat(amf0): serde rewrite

lennartkloock:
- Rewrite scuffle-amf0 with serde support
  - Adheres to these [conventions](https://serde.rs/conventions.html)

CLOUD-91


crates/amf0/src/de.rs
```diff
@@ -0,0 +1,1065 @@
...
+        let marker = self.reader.read_u8()?;
+        let marker = Amf0Marker::from_u8(marker).ok_or(Amf0Error::UnknownMarker(marker))?;
+
+        self.reader.seek_relative(-1)?;
+
+        match marker {
+            Amf0Marker::Boolean => self.deserialize_bool(visitor),
+            Amf0Marker::Number | Amf0Marker::Date => self.deserialize_f64(visitor),
+            Amf0Marker::String | Amf0Marker::LongString | Amf0Marker::XmlDocument => self.deserialize_str(visitor),
+            Amf0Marker::Null | Amf0Marker::Undefined => self.deserialize_unit(visitor),
```
TroyKomodo, line 132: this should be `self.deserialize_none` for optional values

crates/amf0/src/de.rs
```diff
@@ -0,0 +1,1065 @@
...
+            Amf0Marker::Object | Amf0Marker::TypedObject | Amf0Marker::EcmaArray => self.deserialize_map(visitor),
```
TroyKomodo, line 133: are is the `EchmaArray` not just an array with any values? where as `StrictArray` is all the same types?

crates/amf0/src/de.rs
```diff
@@ -0,0 +1,1065 @@
...
+        let value = self.read_number()?;
+        visitor.visit_f64(value)
+    }
+
+    fn deserialize_char<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: serde::de::Visitor<'de>,
+    {
+        let value = self.read_number()?;
+        visitor.visit_char(value as u8 as char)
```
TroyKomodo, line 226: we should check if the value can be safely converted to a char without loss

crates/amf0/src/de.rs
```diff
@@ -0,0 +1,1065 @@
...
+
+impl Deserializer {
+    /// Create a new deserializer from a reader.
+    pub fn new(bytes: Bytes) -> Self {
+        Deserializer {
+            reader: io::Cursor::new(bytes),
+        }
+    }
+
+    fn expect_marker(&mut self, expect: &'static [Amf0Marker]) -> Result<(), Amf0Error> {
```
TroyKomodo, line 36: perhaps return the marker that matched?

crates/amf0/src/de.rs
```diff
@@ -0,0 +1,1065 @@
...
+        V: serde::de::Visitor<'de>,
+    {
+        self.deserialize_tuple(len, visitor)
+    }
+
+    fn deserialize_map<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: serde::de::Visitor<'de>,
+    {
+        let marker = self.reader.read_u8()?;
```
TroyKomodo, line 349: if we convert the expect_marker to return the marker that matched, then this becomes easier to write

crates/amf0/src/de.rs
lennartkloock: There is no function with that name.

crates/amf0/src/de.rs
lennartkloock: EcmaArrays are encoded as objects with a known length. That means the only difference to StrictArrays is that the index/key is a string instead of a number. Both can contain differently typed values.

crates/amf0/src/de.rs
lennartkloock: How would that look? This conversion floors the f64 to the nearest u8. u8 can be safely converted to char.

crates/rtmp/src/command_messages/on_status/mod.rs
```diff
@@ -18,42 +18,39 @@ pub mod writer;
 #[derive(Debug, Clone, PartialEq)]
 pub struct OnStatus<'a> {
     /// The status code.
-    ///
-    /// See the [`codes`] module for common status codes.
-    pub code: Cow<'a, str>,
```
philipch07, line 23: Perhaps link to the ```[`OnStatusCode`]``` enum for easier reference in an ide when peeking the definition for the struct.

crates/amf0/src/ser.rs
```diff
@@ -0,0 +1,978 @@
...
+    type Ok = ();
+
+    fn serialize_element<T>(&mut self, value: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + serde::Serialize,
+    {
+        value.serialize(&mut **self)
+    }
+
+    fn end(self) -> Result<Self::Ok, Self::Error> {
```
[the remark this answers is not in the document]
lennartkloock: https://docs.rs/serde/latest/serde/ser/trait.SerializeTuple.html#tymethod.end

crates/amf0/src/ser.rs
```diff
@@ -0,0 +1,978 @@
...
+    type Ok = ();
+    type SerializeMap = Impossible<Self::Ok, Self::Error>;
+    type SerializeSeq = Impossible<Self::Ok, Self::Error>;
+    type SerializeStruct = Impossible<Self::Ok, Self::Error>;
+    type SerializeStructVariant = Impossible<Self::Ok, Self::Error>;
+    type SerializeTuple = Impossible<Self::Ok, Self::Error>;
+    type SerializeTupleStruct = Impossible<Self::Ok, Self::Error>;
+    type SerializeTupleVariant = Impossible<Self::Ok, Self::Error>;
+
+    fn serialize_bool(self, _v: bool) -> Result<Self::Ok, Self::Error> {
```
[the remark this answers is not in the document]
lennartkloock: We have to implement it this way because of how serde works.

crates/amf0/src/de.rs
```diff
@@ -0,0 +1,1065 @@
...
+        self.expect_marker(&[Amf0Marker::Boolean])?;
+        let value = self.reader.read_u8()?;
+        visitor.visit_bool(value != 0)
+    }
+
+    fn deserialize_i8<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: serde::de::Visitor<'de>,
+    {
+        self.deserialize_i64(visitor)
```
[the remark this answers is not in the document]
lennartkloock: I try to avoid duplicate code. It's the visitor's responsibility to cast the visited value into the right type.

crates/amf0/src/ser.rs
```diff
@@ -0,0 +1,948 @@
...
+}
+
+impl<W> Serializer<W> {
+    /// Create a new AMF0 serializer from a writer.
+    pub fn new(writer: W) -> Self {
+        Self {
+            encoder: Amf0Encoder::new(writer),
+        }
+    }
+}
```
TroyKomodo, line 42: Does this need to be a new type? can we not just impl serializer for `&Mut amf0Encoder`?

crates/amf0/src/decoder.rs
```diff
@@ -332,3 +332,69 @@ impl<'de, R: ZeroCopyReader<'de>> Iterator for Amf0DecoderStream<'_, 'de, R> {
 }
 
 impl<'de, R: ZeroCopyReader<'de>> std::iter::FusedIterator for Amf0DecoderStream<'_, 'de, R> {}
```
TroyKomodo, line 334: I think this trait should also be applied to the deserialize version

TroyKomodo (approved):

TroyKomodo:
?brawl merge

lennartkloock:
?brawl retry

lennartkloock:
?brawl retry

lennartkloock:
?brawl merge

Shouldn't have done a retry

lennartkloock:
?brawl merge

PR #432, opened 2025-04-21
fix clippy warning from nightly

TroyKomodo:
Nightly got bumped and caused an update in clippy. 
I suspect in 2024 edition the behaviour around the pin! macro changed because I believe you used to require the variable be mut before using the macro.


lennartkloock (approved):

TroyKomodo:
?brawl merge

PR #441, opened 2025-04-27
Add deserializer for amf0 value and fix amf0 visitor fns

philipch07:


crates/amf0/src/value.rs
```diff
@@ -280,12 +280,120 @@ impl serde::ser::Serialize for Amf0Value<'_> {
...
+                    } else if i >= 0 {
+                        visitor.visit_u64(i as u64)
+                    } else {
+                        visitor.visit_f64(num)
+                    };
+                    result
+                } else {
+                    visitor.visit_f64(num)
+                }
+            }
```
TroyKomodo, line 329: this should be visit_f64 always since thats all that we have.

crates/amf0/src/value.rs
```diff
@@ -280,12 +280,120 @@ impl serde::ser::Serialize for Amf0Value<'_> {
...
 }
 
+#[cfg(feature = "serde")]
+#[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
+impl<'de, 'a: 'de> serde::Deserializer<'de> for &'a Amf0Value<'de> {
+    type Error = Amf0Error;
+
+    serde::forward_to_deserialize_any! {
+        // need to handle these number types from i8 thru f64; try to deserialze them as their original type then forward as f64
+        bool i8 i16 i32 i64 u8 u16 u32 u64 f32 f64 char str string unit
```
TroyKomodo, line 290: Here where we have `i8 i16 i32 i64 u8 u16 u32 u64 f32` we should convert try convert f64 to these types if it fits in that type (see `num_traits` crate)

crates/amf0/src/value.rs
```diff
@@ -280,12 +280,120 @@ impl serde::ser::Serialize for Amf0Value<'_> {
...
+        option seq bytes byte_buf map unit_struct newtype_struct tuple tuple_struct
+        struct enum identifier ignored_any
```
TroyKomodo, line 292: These should also be handled in the same way the deserializer handles them

crates/amf0/src/de/mod.rs
```diff
@@ -77,21 +77,24 @@ where
     where
         V: serde::de::Visitor<'de>,
     {
-        self.deserialize_i64(visitor)
+        let value = self.decode_number()?;
```
TroyKomodo, line 80: do `peek_marker` and if the marker is a number do decode number otherwise forward to `deserialize_any`. 

If it is a number try convert the number into a i8; use num_traits if that conversion fails just fall back to visit_f64

TroyKomodo (approved):

TroyKomodo (changes requested):
good first steps

crates/amf0/src/de/mod.rs
```diff
@@ -77,73 +77,110 @@ where
     where
         V: serde::de::Visitor<'de>,
     {
-        self.deserialize_i64(visitor)
+        if let Some(Amf0Marker::Number) = self.next_marker {
+            let value = self.decode_number()?;
+            return visitor.visit_i8(value as i8);
+        }
+        self.deserialize_any(visitor)
```
TroyKomodo, line 84: this isnt correct yet, because we need to make sure the value can fit in the i8

crates/amf0/src/de/mod.rs
```diff
@@ -42,12 +42,57 @@ where
     Ok(value)
 }
 
+#[cfg(feature = "serde")]
+#[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
```
TroyKomodo, line 46: We dont need `doc(cfg)` because we have a crate feature `#![cfg_attr(docsrs, feature(doc_auto_cfg))]`

crates/amf0/src/value.rs
```diff
@@ -1,18 +1,19 @@
...
 
 use scuffle_bytes_util::StringCow;
+use serde::de::IntoDeserializer;
 
 use crate::Amf0Error;
 use crate::encoder::Amf0Encoder;
 
 /// Represents any AMF0 object.
-pub type Amf0Object<'a> = HashMap<StringCow<'a>, Amf0Value<'a>>;
+pub type Amf0Object<'a> = BTreeMap<StringCow<'a>, Amf0Value<'a>>;
```
TroyKomodo, line 14: We should make this a wrapper instead of a type def.

So right now this is assigned to be a BTreeMap, this is less than ideal because in the future we might want to preserve order strictly (ie an `IndexMap`) and that would be a breaking change. Instead we should make a new structure called `Amf0Object` and then internally we have a `BTreeMap` and then we provide a bunch of helper functions reexporting the BTreeMap's behaviours (`len`, `is_empty`, `iter`, `into_iter`, `get`, `get_mut`, `remove`, `insert`, `from_iter`, etc)

crates/amf0/src/value.rs
```diff
@@ -278,15 +261,173 @@ impl serde::ser::Serialize for Amf0Value<'_> {
...
+}
+
+#[cfg(feature = "serde")]
+macro_rules! impl_deserializer {
+    ($ty:ty) => {
+        impl<'de> serde::Deserializer<'de> for $ty {
+            type Error = Amf0Error;
+
+            serde::forward_to_deserialize_any! {
+                bool f64 char str string unit
```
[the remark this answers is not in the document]
philipch07: Should it just throw an `Amf0::CharNotSupported` error since it isn't supported? Like the following:
```rust
fn deserialize_char<V>(self, _visitor: V) -> Result<V::Value, Self::Error>
where
    V: serde::de::Visitor<'de>,
{
    Err(Amf0Error::CharNotSupported)
}
```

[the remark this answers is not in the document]
philipch07: Oh wait I just saw your next comment

crates/amf0/src/object.rs
```diff
@@ -0,0 +1,1000 @@
+//! This code is modified from <https://github.com/serde-rs/json/blob/v1.0.140/src/map.rs>
```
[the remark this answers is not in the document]
philipch07: I can write some tests to improve the coverage

crates/amf0/src/value.rs
```diff
@@ -278,15 +265,204 @@ impl serde::ser::Serialize for Amf0Value<'_> {
...
+
+macro_rules! impl_deserializer {
+    ($ty:ty) => {
+        impl<'de> serde::Deserializer<'de> for $ty {
+            type Error = Amf0Error;
+
+            serde::forward_to_deserialize_any! {
+                bool f64 char str string unit
+                seq map newtype_struct tuple
+                struct enum ignored_any identifier
```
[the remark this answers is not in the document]
philipch07: I made a ticket for it for now, I might tackle it in the future: CLOUD-123

crates/amf0/src/lib.rs
```diff
@@ -41,16 +41,20 @@
 //!
 //! `SPDX-License-Identifier: MIT OR Apache-2.0`
 #![cfg_attr(all(coverage_nightly, test), feature(coverage_attribute))]
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_auto_cfg, doc_cfg))]
```
TroyKomodo, line 44: no we dont need doc_cfg if we have doc_auto_cfg

crates/amf0/src/object.rs
```diff
@@ -0,0 +1,2486 @@
...
+
+    /// Removes and returns the value corresponding to the key from the map.
+    ///
+    /// Like [`Vec::swap_remove`], the entry is removed by swapping it with the
+    /// last element of the map and popping it off. This perturbs the position
+    /// of what used to be the last element!
+    ///
+    /// [`Vec::swap_remove`]: std::vec::Vec::swap_remove
+    #[cfg(feature = "preserve_order")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "preserve_order")))]
```
TroyKomodo, line 201: we can remove this since we use `doc_auto_cfg`

crates/amf0/src/de/mod.rs
```diff
@@ -42,156 +42,164 @@ where
...
 
     fn deserialize_string<V>(self, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: serde::de::Visitor<'de>,
     {
-        let value = self.decode_string()?;
-        value.into_deserializer().deserialize_string(visitor)
+        if let Amf0Marker::String | Amf0Marker::LongString | Amf0Marker::XmlDocument = self.peek_marker()? {
+            let value = self.decode_string()?;
+            value.into_deserializer().deserialize_string(visitor)
```
TroyKomodo, line 157: Why cant we just call `visitor.visit_string`

crates/amf0/src/de/mod.rs
```diff
@@ -42,156 +42,164 @@ where
...
 
     fn deserialize_str<V>(self, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: serde::de::Visitor<'de>,
     {
-        let value = self.decode_string()?;
-        value.into_deserializer().deserialize_str(visitor)
+        if let Amf0Marker::String | Amf0Marker::LongString | Amf0Marker::XmlDocument = self.peek_marker()? {
+            let value = self.decode_string()?;
+            visitor.visit_string(value.to_string())
```
TroyKomodo, line 169: In this case we should call `visitor.visit_str` if we dont have an owned.

crates/amf0/src/de/mod.rs
```diff
@@ -228,52 +240,61 @@ where
...
     fn deserialize_tuple_struct<V>(self, _name: &'static str, len: usize, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: serde::de::Visitor<'de>,
     {
-        self.deserialize_tuple(len, visitor)
+        if let Amf0Marker::StrictArray = self.peek_marker()? {
+            self.deserialize_tuple(len, visitor)
+        } else {
+            self.deserialize_any(visitor)
+        }
```
TroyKomodo, line 279: why even bother jsut call `self.deserialize_tuple`

TroyKomodo (changes requested):

crates/amf0/src/de/mod.rs
```diff
@@ -228,31 +230,32 @@ where
...
-            return Err(Amf0Error::WrongArrayLength {
-                expected: len,
-                got: size,
-            });
+            visitor.visit_seq(StrictArray {
+                de: self,
+                remaining: size,
+            })
+        } else {
+            self.deserialize_any(visitor)
```
TroyKomodo, line 257: this seems to do the same thing as `deserialize_seq`

crates/amf0/src/de/mod.rs
```diff
@@ -228,31 +230,32 @@ where
...
+        }
     }
 
-    fn deserialize_tuple<V>(self, len: usize, visitor: V) -> Result<V::Value, Self::Error>
+    fn deserialize_tuple<V>(self, _len: usize, visitor: V) -> Result<V::Value, Self::Error>
```
TroyKomodo, line 245: forward to any

crates/amf0/src/de/mod.rs
```diff
@@ -228,31 +230,32 @@ where
...
+            self.deserialize_any(visitor)
         }
-
-        visitor.visit_seq(StrictArray {
-            de: self,
-            remaining: size,
-        })
     }
 
     fn deserialize_tuple_struct<V>(self, _name: &'static str, len: usize, visitor: V) -> Result<V::Value, Self::Error>
```
TroyKomodo, line 252: forward to any

crates/amf0/src/de/mod.rs
```diff
@@ -305,15 +312,12 @@ where
...
-        let s = self.decode_string()?;
-        s.into_deserializer().deserialize_identifier(visitor)
-    }
-
-    fn deserialize_ignored_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
-    where
-        V: serde::de::Visitor<'de>,
-    {
-        self.deserialize_any(visitor)
+        if let Amf0Marker::String | Amf0Marker::LongString = self.peek_marker()? {
```
TroyKomodo, line 306: forward to any

crates/amf0/src/de/mod.rs
```diff
@@ -202,8 +200,12 @@ where
...
-        visitor.visit_unit()
+        if let Amf0Marker::Null | Amf0Marker::Undefined = self.peek_marker()? {
+            self.decode_null()?;
+            visitor.visit_unit()
+        } else {
+            self.deserialize_any(visitor)
+        }
     }
 
     fn deserialize_unit_struct<V>(self, _name: &'static str, visitor: V) -> Result<V::Value, Self::Error>
```
TroyKomodo, line 211: forward to any

crates/amf0/src/de/mod.rs
```diff
@@ -202,8 +200,12 @@ where
     where
```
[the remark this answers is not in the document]
philipch07: Removing this causes the `optional()` test to fail. My current understanding is that forwarding to any doesn't handle the null properly so it keeps looping over and over causing a stack overflow.

PR #450, opened 2025-05-05
Fix semver checks bad unwrap

philipch07:
Fixes a bad unwrap as seen in: https://github.com/ScuffleCloud/scuffle/actions/runs/14829284811/job/41627342997#step:8:282

~~Currently based off of Troy's tinc branch to make sure it works.~~ I'm pretty sure it works now

### DIFF
--- a/dev-tools/xtask/src/cmd/semver_checks/mod.rs
+++ b/dev-tools/xtask/src/cmd/semver_checks/mod.rs
@@ -182,15 +182,26 @@ impl SemverChecks {
                         // need new line to allow for bullet list formatting
                         description.push("".to_string());
 
+                        // at this point, we begin parsing the
                         let file_loc = desc_trimmed
                             .split_whitespace()
-                            .last()
-                            .unwrap()
-                            .strip_prefix("/home/runner/work/scuffle/scuffle/")
-                            .unwrap()
-                            .replace(":", "#L");
+                            .last() // get the file location string (the last string in the line)
+                            .unwrap();
 
-                        description.push(format!("- {scuffle_commit_url}/{file_loc}"));
+                        // remove the prefix if it exists, otherwise use the original string
+                        // for reference, the Some case would be something like:
+                        // field stdout of struct CompileOutput, previously in file "/home/runner/work/scuffle/scuffle/..."
+                        // but the other case would be something like:
+                        // "feature prettyplease in the package's Cargo.toml"
+                        match file_loc.strip_prefix("/home/runner/work/scuffle/scuffle/") {
+                            Some(stripped) => {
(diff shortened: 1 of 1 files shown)


dev-tools/xtask/src/cmd/semver_checks/mod.rs
```diff
@@ -182,15 +182,26 @@ impl SemverChecks {
...
-                        description.push(format!("- {scuffle_commit_url}/{file_loc}"));
+                            .last() // get the file location string (the last string in the line)
+                            .unwrap();
+
+                        // remove the prefix if it exists, otherwise use the original string
+                        // for reference, the Some case would be something like:
+                        // field stdout of struct CompileOutput, previously in file "/home/runner/work/scuffle/scuffle/..."
+                        // but the other case would be something like:
+                        // "feature prettyplease in the package's Cargo.toml"
+                        match file_loc.strip_prefix("/home/runner/work/scuffle/scuffle/") {
```
TroyKomodo, line 196: is it not possible to make this less specialized like make it work within the workspace root or something.

TroyKomodo (changes requested):

dev-tools/xtask/src/cmd/semver_checks/mod.rs
philipch07: Good idea, I'll check if there's an env variable we have for it.

TroyKomodo: No, we should just use cargo to find the root directory using cargo metadata.

philipch07: Oh whoops. I'll try that now

dev-tools/xtask/src/cmd/semver_checks/mod.rs
```diff
@@ -182,15 +182,26 @@ impl SemverChecks {
...
-                        description.push(format!("- {scuffle_commit_url}/{file_loc}"));
+                            .last() // get the file location string (the last string in the line)
+                            .unwrap();
+
+                        // remove the prefix if it exists, otherwise use the original string
+                        // for reference, the Some case would be something like:
+                        // field stdout of struct CompileOutput, previously in file "/home/runner/work/scuffle/scuffle/..."
+                        // but the other case would be something like:
+                        // "feature prettyplease in the package's Cargo.toml"
+                        match file_loc.strip_prefix(&(current_metadata.workspace_root.to_string() + "/")) {
```
TroyKomodo, line 196: use the format macro instead. `format!("{}/", current_metadata.workspace_root)`

TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #456, opened 2025-05-13
remove workspace-deps

TroyKomodo:
@SimaoMoreira5228 's work for moving away from workspace deps.

The reason being is that we want to be able to select versions of workspace members in our crates more precisely. I added some fixes to post compile to fix build caching.

### DIFF
--- a/crates/amf0/Cargo.toml
+++ b/crates/amf0/Cargo.toml
@@ -18,7 +18,7 @@ num-traits = "0.2"
 num-derive = "0.4"
 thiserror = "2.0"
 bytestring = "1.4.0"
-scuffle-bytes-util = { workspace = true }
+scuffle-bytes-util = "0.1.1"
 scuffle-workspace-hack.workspace = true
 
 [dev-dependencies]
--- a/crates/postcompile/src/lib.rs
+++ b/crates/postcompile/src/lib.rs
@@ -204,8 +204,10 @@ impl std::fmt::Display for CompileOutput {
 fn cargo(config: &Config, manifest_path: &Path, subcommand: &str) -> Command {
     let mut program = Command::new(std::env::var_os("CARGO").unwrap_or_else(|| "cargo".into()));
     program.arg(subcommand);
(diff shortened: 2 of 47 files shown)


crates/amf0/Cargo.toml
```diff
@@ -18,7 +18,7 @@ num-traits = "0.2"
 num-derive = "0.4"
 thiserror = "2.0"
 bytestring = "1.4.0"
-scuffle-bytes-util = { workspace = true }
+scuffle-bytes-util = "0.1.1"
```
philipch07, line 21: Should we avoiding pinning to the patch version?

crates/postcompile/src/lib.rs
```diff
@@ -376,12 +398,16 @@ pub fn compile_custom(tokens: impl std::fmt::Display, config: &Config) -> std::i
...
-    // We set this env variable so that this compiler can accept nightly options.
+    // We set this env variable so that this compiler can accept nightly options.)
     program.env("RUSTC_BOOTSTRAP", "1");
     program.arg("--").arg("-Zunpretty=expanded");
 
+    let start = std::time::Instant::now();
+
     let output = program.output().unwrap();
 
+    dbg!(start.elapsed());
```
philipch07, line 409: Stray dbg?

crates/postcompile/src/lib.rs
```diff
@@ -418,7 +444,9 @@ pub fn compile_custom(tokens: impl std::fmt::Display, config: &Config) -> std::i
             program.arg("--no-run");
         }
 
+        let start = std::time::Instant::now();
         let comp_output = program.output().unwrap();
+        dbg!(start.elapsed());
```
philipch07, line 449: Stray dbg?

crates/amf0/Cargo.toml
TroyKomodo: no for our own deps we can just specify a patch

philipch07 (approved):

TroyKomodo:
?brawl merge

PR #457, opened 2025-05-13
Do semver against crates.io

SimaoMoreira5228:
Update semver check to use latest version from crates.io instead of main branch.


philipch07:
I'll fix the second error output in a little bit.

PR #459, opened 2025-05-14
Simao/semver against crates io

philipch07:
test (idk why it made a new branch on scuffle instead of using the existing one but oh well...)

### DIFF
--- a/dev-tools/xtask/Cargo.toml
+++ b/dev-tools/xtask/Cargo.toml
@@ -22,4 +22,5 @@ semver = "1.0.26"
 serde = "1"
 serde_derive = "1"
 serde_json = "1.0"
-toml_edit = { features = ["serde"], version = "0.22" }
+toml_edit = { version = "0.22", features = ["serde"] }
+reqwest ={ version = "0.12", features = ["blocking", "json"] }
--- a/dev-tools/xtask/Cargo.toml
+++ b/dev-tools/xtask/Cargo.toml
@@ -16,7 +16,6 @@ anyhow = "1.0"
 cargo_metadata = "0.19.1"
 clap = { features = ["derive", "env"], version = "4.5.23" }
 next_version = "0.2.24"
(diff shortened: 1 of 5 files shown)


dev-tools/xtask/src/cmd/semver_checks/mod.rs
```diff
@@ -62,35 +53,51 @@ impl SemverChecks {
...
-            "--all-features",
-        ];
+        // Safety: this is safe because we are setting the environment variable for the current process.
+        unsafe {
+            std::env::set_var("RUSTC_BOOTSTRAP", "1");
+            std::env::set_var("RUSTDOCFLAGS", "-Z unstable-options --output-format json");
+        }
+
+        // make the rustdoc json
+        let rustdoc_args = vec!["doc", "--no-deps", "--all-features"];
```
philipch07, line 63: Need to just go back to `rustdoc` and fix up the args. It would be useful to output each json into `./target/doc` though. @SimaoMoreira5228 if you'd like to work on this feel free as I'll be a little more busy tomorrow most likely. You'll probably have to pull and sync to your branch, but also it may be a little annoying to get the runners to execute, but I think you might be able to just run this locally so you can probably test it that way.

dev-tools/xtask/Cargo.toml
```diff
@@ -16,10 +16,10 @@ anyhow = "1.0"
...
 next_version = "0.2.24"
-os_pipe = "1.2.1"
 regex = "1.11.1"
 semver = "1.0.26"
 serde = "1"
 serde_derive = "1"
 serde_json = "1.0"
-toml_edit = { features = ["serde"], version = "0.22" }
+toml_edit = { version = "0.22", features = ["serde"] }
+reqwest ={ version = "0.12", features = ["blocking", "json"] }
```
TroyKomodo, line 25: still adding request here

PR #460, opened 2025-05-15
fix workspace deps

TroyKomodo:
Cargo is stupid...

So because dev-dependencies are allowed to have cycles (this makes sense since they don't effect the build tree of downstream crates) cargo struggles to find the right order to publish because for some reason each dev-dependency must be on crates.io prior to the publish succeeding (this does not make sense).


TroyKomodo:
?brawl merge

TroyKomodo:
?brawl merge

PR #462, opened 2025-05-15
Set openapi-v3 version in tinc-build

philipch07:
Should fix the crates.io release which failed here: https://github.com/ScuffleCloud/scuffle/actions/runs/15039106693/job/42266458336


philipch07:
?brawl try

TroyKomodo:
?brawl cancel

PR #463, opened 2025-05-15
apply current version to all crates

TroyKomodo:
supersedes #462 to apply the current minimum version to all crates


TroyKomodo:
?brawl merge

PR #473, opened 2025-05-24
release improvements

TroyKomodo:
Improvements to the release process

The new release process is as follows:

After you change or create a new crate, you should commit and then run this.

```
cargo xtask release fix --pr-number <pr-number>
```

to automatically fix any issues with your updates.

You can run the `check` version too if u want to see issues without fixing them.

```
cargo xtask release check [--pr-number <pr-number>]
```

(in this case the pr number is optional) 

In CI the bot will also run and add a comment on the PR listing all package changes (showing API breakage status).

Sometimes you might get a `min-versions error` this happens when you update a package and then depend on those updates in another package. In order to deal with this you must change the version requirement of the dependency to be inclusive of future versions. 

An example.

`Cargo.toml`

```toml
[workspace]
members = [
  "crates/*"
]

resolver = "3"
```

`crates/a/Cargo.toml`

```toml
[package]
name = "a"
version = "0.1.3"

[dependencies]
```

`crates/b/Cargo.toml`

```toml
[package]
name = "b"
version = "0.1.0"

[dependencies]
a = { version = "0.1", path = "../a" }
```

Lets say as part of your PR your diff looks like such

```diff
--- a/crates/a/src/lib.rs
+++ a/crates/a/src/lib.rs
@@ -0,0 +0,0 @@
+ pub fn some_cool_new_function() {
+     // ...
+ }

--- a/crates/b/src/lib.rs
+++ a/crates/b/src/lib.rs
@@ -0,0 +0,0 @@
+ pub fn even_cooler_function() {
+     a::some_cool_new_function()
+ }
(END)
```

Because `b` depends on unreleased changes in `a` you must update `b`'s `Cargo.toml` to reflect this.

`crates/b/Cargo.toml`

```toml
[package]
name = "b"
version = "0.1.0"

[dependencies]
a = { version = ">=0.1.3", path = "../a" }
```

By denoting the `>=0.1.0` in the version specifier you inform the release calculator that `b` depends on `a`'s unreleased changes so when you bump `a`'s version you must insert it into `b`'s dependencies list. 

When a pr is generated (via `cargo xtask release pr`), the `Cargo.toml`s will be updated to the following:

`crates/a/Cargo.toml`

```toml
[package]
name = "a"
version = "0.1.4"

[dependencies]
```

`crates/b/Cargo.toml`

```toml
[package]
name = "b"
version = "0.1.1"

[dependencies]
a = { version = "0.1.4", path = "../a" }
```

## Other changes in this PR

This PR also changes slightly how changelogs work. In the past we always had to add the `change-log-not-required` label to PRs which do not need changelogs. Moving forward changelogs will automatically be calculated based on what packages change and therefore do not need to be provided unless a package which has a changelog has changed.

This PR therefore removes the `change-log-not-required` label as it is now unused.

### DIFF
--- a/.github/labels.yaml
+++ b/.github/labels.yaml
@@ -28,8 +28,6 @@
   color: "d876e3"
 - name: "wontfix"
   color: "ffffff"
-- name: "change-log-not-required"
-  color: "ffffff"
 - name: "release"
   color: "32a852"
 - name: "automated"
--- a/.github/workflows/cargo-update-pr.yaml
+++ b/.github/workflows/cargo-update-pr.yaml
@@ -50,7 +50,7 @@ jobs:
           title: Automated update of Cargo.lock
           body-path: /tmp/pr-body.txt
           branch: automation/cargo-update
-          labels: dependencies,automated,change-log-not-required
+          labels: dependencies,automated
           commit-message: |
             Automated update of Cargo.lock
 


philipch07:
Would it be reasonable to change the usage to be something like: `cargo xtask release check/fix <pr-number>`? The difference here is that we wouldn't have to specify the flag `--pr-number`, and instead directly pass the number. I'm just curious about this since I'm not sure how many other flags we may have to set on `check` or `fix` (if any).

TroyKomodo:
> Would it be reasonable to change the usage to be something like: `cargo xtask release check/fix <pr-number>`? The difference here is that we wouldn't have to specify the flag `--pr-number`, and instead directly pass the number. I'm just curious about this since I'm not sure how many other flags we may have to set on `check` or `fix` (if any).

There are other flags

`cargo xtask release check`
```
Check everything is correct for a merge

Usage: cargo xtask release check [OPTIONS]

Options:
  -n, --pr-number <PR_NUMBER>      The pull request number
      --base-branch <BASE_BRANCH>  The base branch to compare against to determine if something has changed [default: origin/main]
      --all                        Check everything, even if there are no changes from this branch to the base branch
  -p, --package <PACKAGES>         Packages to include in the check by default all packages are included
      --allow-dirty                Allow to run when there are uncomitted changes
      --stdout-markdown            Output markdown to stdout (used for CI to generate PR comments / Summaries)
      --exit-status                Return a non-zero exit status at the end if a check failed
      --concurrency <CONCURRENCY>  [default: 128]
  -h, --help                       Print help
```

`cargo xtask release check`
```
Fix all problems

Usage: cargo xtask release fix [OPTIONS] --pr-number <PR_NUMBER>

Options:
  -n, --pr-number <PR_NUMBER>      The pull request number
      --base-branch <BASE_BRANCH>  The base branch to compare against to determine if something has changed [default: origin/main]
      --all                        Check everything, even if there are no changes from this branch to the base branch
  -p, --package <PACKAGES>         Packages to include in the check by default all packages are included
      --allow-dirty                Allow to run when there are uncomitted changes
      --concurrency <CONCURRENCY>  [default: 128]
      --author <AUTHORS>           Author to use for the changelog entries
  -h, --help                       Print help
```

dev-tools/xtask/src/cmd/release/publish.rs
```diff
@@ -0,0 +1,209 @@
...
+use super::utils::{GitReleaseArtifact, Package, concurrently, git_workdir_clean};
+use crate::utils::{self, Command, cargo_cmd};
+
+#[derive(Debug, Clone, clap::Parser)]
+pub struct Publish {
+    /// Packages to include in the check
+    /// by default all packages are included
+    #[arg(long = "package", short = 'p')]
+    packages: Vec<String>,
+    /// Allow to run when there are uncomitted changes.
```
philipch07, line 14:
```suggestion
    /// Allow this to run when there are uncommitted changes.
```

dev-tools/xtask/src/cmd/release/publish.rs
```diff
@@ -0,0 +1,209 @@
...
+    #[arg(long)]
+    allow_dirty: bool,
+    // Concurrency to run at
```
philipch07, line 17:
```suggestion
    /// Concurrency to run at. By default, this is the total number of cpus on the host.
```

philipch07 (commented):
Overall looks really solid, I think all my comments are just nits and asking for a couple more comments/explanations for certain sections so it'll be easier for us to maintain in the future since there seems to be a decent number of moving parts.

philipch07:
> > Would it be reasonable to change the usage to be something like: `cargo xtask release check/fix <pr-number>`? The difference here is that we wouldn't have to specify the flag `--pr-number`, and instead directly pass the number. I'm just curious about this since I'm not sure how many other flags we may have to set on `check` or `fix` (if any).
> 
> There are other flags
> 
> `cargo xtask release check`
> 
> ```
> Check everything is correct for a merge
> 
> Usage: cargo xtask release check [OPTIONS]
> 
> Options:
>   -n, --pr-number <PR_NUMBER>      The pull request number
>       --base-branch <BASE_BRANCH>  The base branch to compare against to determine if something has changed [default: origin/main]
>       --all                        Check everything, even if there are no changes from this branch to the base branch
>   -p, --package <PACKAGES>         Packages to include in the check by default all packages are included
>       --allow-dirty                Allow to run when there are uncomitted changes
>       --stdout-markdown            Output markdown to stdout (used for CI to generate PR comments / Summaries)
>       --exit-status                Return a non-zero exit status at the end if a check failed
>       --concurrency <CONCURRENCY>  [default: 128]
>   -h, --help                       Print help
> ```
> 
> `cargo xtask release fix`
> 
> ```
> Fix all problems
> 
> Usage: cargo xtask release fix [OPTIONS] --pr-number <PR_NUMBER>
> 
> Options:
>   -n, --pr-number <PR_NUMBER>      The pull request number
>       --base-branch <BASE_BRANCH>  The base branch to compare against to determine if something has changed [default: origin/main]
>       --all                        Check everything, even if there are no changes from this branch to the base branch
>   -p, --package <PACKAGES>         Packages to include in the check by default all packages are included
>       --allow-dirty                Allow to run when there are uncomitted changes
>       --concurrency <CONCURRENCY>  [default: 128]
>       --author <AUTHORS>           Author to use for the changelog entries
>   -h, --help                       Print help
> ```

I see, that makes sense. Also, the concurrency default is just the number of cpus, is that a variable number in the generated description? I know that it calls `num_cpus::get()` but I just wanted to double check since 128 cpus sounds like a lot of cpus to have as a default (or maybe you were running this in your server).

dev-tools/xtask/src/cmd/release/check.rs
```diff
@@ -0,0 +1,553 @@
...
+                    if stdout.is_empty() {
+                        anyhow::bail!("semver-checks failed\n{stderr}");
+                    } else {
+                        self.set_semver_output(stderr.contains("requires new major version"), stdout.to_owned());
+                    }
+                } else {
+                    self.set_semver_output(false, stdout.to_owned());
+                }
+            } else {
+                tracing::info!("skipping semver-checks due to no published-version");
```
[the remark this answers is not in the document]
TroyKomodo: The package name is added via a span on the function
```rust
#[tracing::instrument(skip_all, fields(package = %self.name))]
```

dev-tools/xtask/Cargo.toml
```diff
@@ -1,6 +1,5 @@
 [package]
 name = "xtask"
-version = "0.1.0"
```
lennartkloock, line 3: why is the version removed here?

TroyKomodo: because its a private crate that will never be published, it doesnt need a version

philipch07 (approved):

TroyKomodo:
?brawl merge

TroyKomodo:
?brawl try

TroyKomodo:
?brawl merge

PR #482, opened 2025-05-31
fix release checks color output

TroyKomodo:
Disables color output for release checks min versions.
Also fixes a slight alignment issue.


TroyKomodo:
?brawl merge

PR #484, opened 2025-06-02
minor fixes

TroyKomodo:
* fixed bug min-versions check for new crates
* cleaned up output for update & check
* fixed changelogs not having authors
* fixed git check to ignore 0-diff-trees
* improved release process to use metadata


TroyKomodo:
?brawl merge

PR #492, opened 2025-06-09
chore: fix new lint

lennartkloock:
Fixes newly added lint https://github.com/rust-lang/rust/pull/138677


philipch07 (approved):
Lgtm assuming you add the changelogs for these

PR #493, opened 2025-06-10
Start with i64::MIN for previous_dts to support negative DTS

zezic:
DTS can be negative when ffmpeg encoder generates packet. If we start from zero, then encoder wrapper may crash at assert: https://github.com/ScuffleCloud/scuffle/blob/main/crates/ffmpeg/src/encoder.rs#L264-L269

### DIFF
--- a/changes.d/pr-493.toml
+++ b/changes.d/pr-493.toml
@@ -1,0 +1,4 @@
+[[scuffle-ffmpeg]]
+category = "fix"
+description = "fixed panic when receiving packet with negative DTS from encoder"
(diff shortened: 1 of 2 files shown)


zezic:
Sure! Fixed clippy warning and added changelog.

TroyKomodo:
This seems fine, do you mind adding a unit test / sample file into the repo so we can ensure we dont regress in the future.

aureateAnatidae:
@zezic Hey, thanks for your work. If it's alright with you, may I handle writing the regression test for your PR?

PR #496, opened 2025-06-14
update fmt style

TroyKomodo:
Update the fmt style to always expand enum struct variants to be multi-line.

Reasoning: if I wanted a single line, I likely would have defined it as a tuple variant.

### DIFF
--- a/changes.d/pr-496.toml
+++ b/changes.d/pr-496.toml
@@ -1,0 +1,14 @@
+[[scuffle-amf0]]
+category = "chore"
+description = "update fmt style"
+authors = ["troykomodo"]
+
+[[scuffle-mp4]]
+category = "chore"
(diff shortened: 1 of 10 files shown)


changes.d/pr-496.toml
```diff
@@ -0,0 +1,14 @@
+[[scuffle-amf0]]
+category = "chore"
+description = "update fmt style"
+authors = ["troykomodo"]
```
lennartkloock, line 4: Isn't the `@` missing here?

lennartkloock (approved):

changes.d/pr-496.toml
TroyKomodo: its auto inserted if missing

TroyKomodo:
?brawl merge

PR #504, opened 2025-07-11
fix: clippy::manual_is_multiple_of

lennartkloock:
New clippy lint [`clippy::manual_is_multiple_of`](https://rust-lang.github.io/rust-clippy/master/index.html#/manual_is_multiple_of)


philipch07 (changes requested):
I think tinc needs a changelog ~~because it depends on tinc build. It probably needs to have that one notation saying that it'll use the new version of tinc build.~~ Besides that, it lgtm

oops, see troy's comment below

lennartkloock:
?brawl merge

PR #515, opened 2025-08-13
Initial svelte repo

lucassshanks:
Leaving a button in the login flow to get into the app

Login related: A state-based login approach seems better so that UX is smoother than routing to login/password-reset or login/password. Can add URL shallow routing to see visual distinction on pages but I don't see common use-cases for needing to hyperlink to a login method

https://echarts.apache.org/ Charts are generated mostly with this so far. I think they're a highly underrated, very nice quality graphing system with nice animations as data is updating. Struggles a bit with very complex use-cases (data zoom slider customizability when navigating around a chart, might need to make our own). Although our use cases sometimes is already hard to represent in most graphs

world map generation: https://geojson-maps.kyd.au/ + mapshaper. Mercator projection in d3. It's already as small as I felt reasonable but I can reduce resolution further whenever we feel necessary. Haven't put it in it's final place yet

Shipping a bunch of stuff in the stream tab that is fairly good placeholder. Colors are a bit off because of themes getting changed but not that important on immediately fixing


lennartkloock (approved):

lucassshanks:
?brawl merge

lennartkloock:
?brawl retry

PR #524, opened 2025-08-26
try fix hevc transcoding

### LINKED ISSUE #509: Scuffle_transmuxer safari playback
### What seems to be the problem?

Im using scuffle_transmuxer to generate fmp4 segments for browser Mediasource playback. The segments, appended to chrome's sourceBuffer play just fine, but won't get accepted by safari. If, however, I omit the audio trafs when building the moof box, video plays just fine in Safari

TroyKomodo:
fixes #509


TroyKomodo:
It seems with flv the codec type is always `hvc1` which is why i changed it here. I am not sure exactly maybe @joetIO

PR #525, opened 2025-08-27
add mergify git merge

TroyKomodo:
setup a git config to automatically resolve merge conflicts in lock files.


philipch07 (approved):
This is cool, does it automatically resolve merge conflicts when rebasing/updating lock files or do we need to call something specific? I'm assuming that since your description said that it's a git config thing then it's automatic but I just wanted to confirm.

TroyKomodo:
> This is cool, does it automatically resolve merge conflicts when rebasing/updating lock files or do we need to call something specific? I'm assuming that since your description said that it's a git config thing then it's automatic but I just wanted to confirm.

This automatically resolves merge conflicts for most files in rust / and lockfiles.

When u encounter a merge conflict in Cargo.lock it will regenerate the lockfile and then stage that.

PR #534, opened 2025-08-27
rename all BUILD to BUILD.bazel

TroyKomodo:
Macos by default has a case insensitive filesystem and folders named `build` collide with files named `BUILD`


TroyKomodo:
?brawl merge

PR #536, opened 2025-08-28
style: 4 spaces instead of 2

lennartkloock:
This PR changes the `dprint` configuration so that for TS, JSON and TOML 4 spaces are used instead of 2


buf.yaml
```diff
@@ -1,5 +1,10 @@
 # For details on buf.yaml configuration, visit https://buf.build/docs/configuration/v2/buf-yaml
 version: v2
+modules:
+  - path: .
+    excludes:
+      - node_modules
+      - website/node_modules
```
TroyKomodo, line 7: maybe we should also add the docs node_modules? infact can it not just ignore all things in git ignore?

TroyKomodo (approved):

TroyKomodo:
?brawl merge

PR #542, opened 2025-08-30
Use dprint instead of prettier 

lucassshanks:
added dprint rules to be the same as the default prettier-svelte support and remove prettier


lucassshanks:
> This is neat, seems like dprint is strictly better than prettier

yea. there's some cases where prettier rules can be more extensive and you can use dprint+prettier but that's w/e

lucassshanks:
?brawl merge

PR #543, opened 2025-08-30
Enhance PR Deploy Comments

SimaoMoreira5228:
Adds merging PR comments that track deployment status across jobs.


SimaoMoreira5228:
It doesn’t update live during the deployment, the comment only gets updated once the job finishes. I agree with capitalizing the table headers and deployment names, as well as replacing the status strings with ✅ / ❌, so I’ll update that.

Regarding the URL, I personally prefer keeping it visible instead of hiding it behind text, but if Troy prefers it hyperlinked, I’m fine making that change.

TroyKomodo:
Yeah i think this is a fine change, i do wish that we could somehow make it update as each one deploys

TroyKomodo:
I think it would be better if we put all the preview jobs into a single workflow file and then include the the pr comment in that PR.

SimaoMoreira5228:
i guess for live update we could do something like a watcher, but i'm not sure how can that be done

.github/workflows/ci.yaml
```diff
@@ -75,60 +75,25 @@ jobs:
...
+    if: ${{ fromJson(needs.prep.outputs.prep).dashboard != null || fromJson(needs.prep.outputs.prep).docs != null || fromJson(needs.prep.outputs.prep).rustdoc != null }}
+    uses: ./.github/workflows/ci-build-previews.yaml
     permissions:
       pull-requests: write
     with:
-      artifact_name: "${{ fromJson(needs.prep.outputs.prep).rustdoc.artifact_name }}"
-      commit_sha: "${{ fromJson(needs.prep.outputs.prep).rustdoc.commit_sha }}"
-      pr_number: "${{ fromJson(needs.prep.outputs.prep).rustdoc.pr_number }}"
-      deploy_docs: "${{ fromJson(needs.prep.outputs.prep).rustdoc.deploy_docs }}"
+      prep_output: "${{ needs.prep.outputs.prep }}"
```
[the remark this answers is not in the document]
SimaoMoreira5228: Do you mean the JSON that the prep script emits?  i.e. the dashboard/docs/rustdoc fields and their keys, should be adjusted so `ci-build-previews` can read them?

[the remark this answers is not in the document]
TroyKomodo: yes

SimaoMoreira5228:
i changed the prep script to return null if the targets do not exist, should i change something else there?

.github/workflows/ci-build-previews.yaml
TroyKomodo, line 1: this file should be called `ci-preview.yaml`

.github/workflows/ci-build-previews.yaml
```diff
@@ -0,0 +1,329 @@
+name: ci - build previews
+
+on:
+  workflow_call:
+    inputs:
+      prep_output:
+        type: string
+        required: true
```
TroyKomodo, line 8: we should break this down into each part too

TroyKomodo (changes requested):

.github/workflows/ci-preview.yaml
```diff
@@ -0,0 +1,342 @@
...
+        run: |
+          pwd=$(pwd)
+          cd ./target-bazel/bin/cloud/dashboard/build
+          tar -cf - * | zstd -o "$pwd/app-scuffle-cloud.tar.zst"
+
+      - name: Upload
+        if: ${{ env.ARTIFACT_NAME != '' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.ARTIFACT_NAME }}
```
TroyKomodo, line 74: Lets just always upload this and call it `ci-preview-rustdoc` `ci-preview-docs` `ci-preview-dashboard` You also will need to update the `release.yaml` file to use the `ci-preview-rustdoc` too

.github/workflows/ci-preview.yaml
```diff
@@ -0,0 +1,336 @@
...
+          body-path: body.md
+
+      - name: Update comment (if found)
+        if: ${{ steps.find_comment.outputs.comment-id != '' }}
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          comment-id: ${{ steps.find_comment.outputs.comment-id }}
+          issue-number: ${{ steps.build_body.outputs.PR }}
+          body-path: body.md
+          edit-mode: replace
```
TroyKomodo, line 336: Why do you have to do these in 3 steps dont the actions typically find / upate the comment for you?

SimaoMoreira5228: its because the comment needs to be created if it doesn't exist, so i need to check if it exists first and then create or update it

TroyKomodo: the previous ones created and updated in a single step? Did u change the comment action?

SimaoMoreira5228: i did change it to peter-evans/find-comment@v3 and peter-evans/create-or-update-comment@v4

SimaoMoreira5228: but i just changed it to the thollander one

.github/scripts/ci-prep.py
```diff
@@ -129,37 +112,30 @@ def deploy_docs() -> bool:
...
-        pr_number=pr_number(),
-        commit_sha=commit_sha(),
+def create_previews() -> Optional[Preview]:
+    has_rustdoc = (
+        os.path.exists("docs")
+        or os.path.exists("target-bazel/bin/docs")
+        or os.path.exists("docs/rustdoc")
     )
+    has_docs = os.path.exists("cloud/docs") or os.path.exists("docs")
+    has_dashboard = os.path.exists("cloud/dashboard")
```
TroyKomodo, line 126: Im confuse by this, why do we check these paths?

SimaoMoreira5228: To check whether the targets are present, either as source dirs or expected build output.

TroyKomodo: I dont think that makes sense or does anything since this is the prep action script it runs before any targets are made and also those paths will always exist because the repo is checked out.

SimaoMoreira5228: i wasn't sure what to put there so i decided to go with the source and target paths

SimaoMoreira5228: i can take a look, it will just take a little bit of time to ajust to bazel and starlark

PR #553, opened 2025-09-03
feat(dashboard): auth

lennartkloock:
This PR adds frontend code that handles the authentication flow with the core service.
This does not implement the UI and only provides an easy-to-use frontend API that can be called from the UI.
This PR also disables some code elements that would no longer work now and need some fixing (`useUser()` for exmaple).


lennartkloock:
?brawl merge

PR #555, opened 2025-09-05
add bazel-diff tool

TroyKomodo:
This tool will allow us to in the future only build / test stuff that has actually changed.


.bazelrc
```diff
@@ -98,6 +98,9 @@ common --execution_log_compact_file=.bazel_compact_execution_log
 # Disable host env from polluting the build
 common --incompatible_strict_action_env
 
+# Use a hermetic JDK
+common --java_runtime_version=remotejdk_21
```
philipch07, line 102: Is this using 21 for any particular reason over 24?

TroyKomodo: i copied this line from the https://github.com/Tinder/bazel-diff/blob/2a1c8fa8fc9f666108eb0bfb489f39b7981212ba/.bazelrc#L7

TroyKomodo: I didnt try 24, but i dont really care to since 21 works fine.

SimaoMoreira5228: java 21 is lts (so its more stable) and 24 is not and the support for it will end in September 2025

TroyKomodo: It also seems that java 21 is an LTS release where as java 24 loses support `JDK 24 will receive updates under these terms, until September 2025, when it will be superseded by JDK 25.` today?

philipch07: > java 21 is lts (so its more stable) and 24 is not and the support for it will end in September 2025

Java 21 isn't necessarily more stable just because it's LTS.

> It also seems that java 21 is an LTS release where as java 24 loses support `JDK 24 will receive updates under these terms, until September 2025, when it will be superseded by JDK 25.` today?

JDK 25 comes out on 9/15 from OpenJDK, which will also be an LTS.

PR #556, opened 2025-09-06
expose rustc flags

TroyKomodo:
adds some options to expose rustc flags / rustc env

cc @lennartkloock


TroyKomodo:
?brawl merge

PR #557, opened 2025-09-06
improve tool performance

TroyKomodo:
This PR improves some tools and allows them to be cached avoiding a costly bazel overhead.


philipch07 (approved):
This is nice, but is there a way to make the rustdoc preview stage any faster? It seems to be the longest stage by far at the moment.

TroyKomodo:
> This is nice, but is there a way to make the rustdoc preview stage any faster? It seems to be the longest stage by far at the moment.

Not really, most of that time is spent when vercel is uploading I think.

TroyKomodo:
?brawl merge

PR #571, opened 2025-09-10
style: deny `clippy::mod_module_files`

lennartkloock:
This PR transitions all crates to the newer self named module style. The `mod.rs` style was replaced with Rust Edition 2018 and the book refers to it as the "older" style.

- https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs
- https://doc.rust-lang.org/reference/items/modules.html#r-items.mod.outlined.search-mod
- https://doc.rust-lang.org/book/ch07-05-separating-modules-into-different-files.html
- https://rust-lang.github.io/rust-clippy/master/index.html#mod_module_files

This might be a quite controversial change. So please comment your opinion.


TroyKomodo:
I think it's fine. I am used to the old style even though I joined after the new style had already become the default. I'm not exactly sure why the rust team decided to change it.

lennartkloock:
> I think it's fine. I am used to the old style even though I joined after the new style had already become the default. I'm not exactly sure why the rust team decided to change it.

I think they changed it because it has some disadvantages to have multiple files called `mod.rs` in your project.
One reason I prefer the new style is because whenever you would add new submodules in the old style, you always had to move the existing module file to a subfolder and rename it to `mod.rs`. With the new style that is no longer needed.

lennartkloock:
?brawl merge

PR #574, opened 2025-09-11
WIP Hookup login flow

lucassshanks:
This PR implements the google auth flow and adds a logout button inside the dashboard

Not entirely sure what would look the best with the layout ex. /login/password vs just /password on the corresponding pages. I think defaulting at login/ and password/ and two-factor/ looks nice though so if you finish at /password you naturally get routed to /two-factor instead. Ultimately not that important

### DIFF
--- a/cloud/dashboard/src/lib/auth/googleAuth.svelte.ts
+++ b/cloud/dashboard/src/lib/auth/googleAuth.svelte.ts
@@ -38,14 +38,7 @@ async function completeGoogleLogin(code: string, state: string): Promise<void> {
         const response = await call.response;
         console.log("Google completion response:", response);
 
-        // I wonder, is it possible to persist somewhere in authState that the login validated successfully
-        // But still needs 2fa?
-        // So if you route to /two-factor you can still access the page with persisted data
-        // I want to route the user to the two-factor page here without needing to pass
-        // information from this completeLoginWithGoogle response ideally?
-        // So I can correctly manage permissioning the two-factor page
-        // It would also be nice if you refresh the two-factor page you're still "half" logged-in
-        // I guess we can do that without decrypting the token
+        // TODO: Implement /two-factor here
 
         // if (response.newUserSessionToken?.sessionMfaPending) {
         //     // Redirect to MFA page


cloud/dashboard/src/lib/auth/googleAuth.svelte.ts
```diff
@@ -0,0 +1,117 @@
...
+    const auth = useAuth();
+
+    const device = await auth.getDeviceOrInit();
+
+    const call = sessionsServiceClient.loginWithGoogle({ device });
+    const status = await call.status;
+
+    if (status.code === "OK") {
+        const response = await call.response;
+        window.location.href = response.authorizationUrl;
```
lennartkloock, line 17: Use Svelte's `goto` instead of this

cloud/dashboard/src/lib/auth/googleAuth.svelte.ts
```diff
@@ -0,0 +1,117 @@
...
+        // So if you route to /two-factor you can still access the page with persisted data
+        // I want to route the user to the two-factor page here without needing to pass
+        // information from this completeLoginWithGoogle response ideally?
+        // So I can correctly manage permissioning the two-factor page
+        // It would also be nice if you refresh the two-factor page you're still "half" logged-in
+        // I guess we can do that without decrypting the token
+
+        // if (response.newUserSessionToken?.sessionMfaPending) {
+        //     // Redirect to MFA page
+        //     window.location.href = "/mfa";
```
lennartkloock, line 55: Same as above

cloud/dashboard/src/lib/auth/googleAuth.svelte.ts
```diff
@@ -0,0 +1,117 @@
...
+        //     return;
+        // }
+
+        if (response.newUserSessionToken) {
+            await auth.handleNewUserSessionToken(response.newUserSessionToken);
+
+            // This should route to whatever our base page is after logging in.
+            // Eventually some dashboard after it's made? Or localhost should throw
+            // redirect to localhost:5173/dashboard after logged in
+            // window.location.href = "/dashboard";
```
lennartkloock, line 65: Same as above

cloud/dashboard/src/lib/auth.svelte.ts
```diff
@@ -126,7 +176,9 @@ function saveDeviceKey(keypair: CryptoKeyPair) {
 
 export type DeviceKeypairState = null | { state: "loading" } | { state: "loaded"; data: CryptoKeyPair | null };
 
-export function authState() {
+let authInstance: ReturnType<typeof createAuthState> | null = null;
+
+export function createAuthState() {
```
lennartkloock, line 181: Why this?

cloud/dashboard/src/lib/auth.svelte.ts
```diff
@@ -277,10 +375,17 @@ async function loadUser(state: AuthState<UserSessionToken>): Promise<AuthState<U
...
         return { state: "error", error: status.detail };
     }
 }
+
+export function useAuth() {
+    if (!authInstance) {
+        authInstance = createAuthState();
+    }
+    return authInstance;
+}
```
lennartkloock, line 391: And this?

cloud/dashboard/src/routes/+layout.ts
```diff
@@ -1 +1,13 @@
+import { useAuth } from "$lib/auth.svelte";
+
 export const ssr = false;
+
+export function load() {
+    const auth = useAuth();
+
+    if (typeof window !== "undefined") {
+        auth.initialize();
```
lennartkloock, line 9: Why did you move this call into the load function? Doesn't that mean it gets called every time someone navigates to another page instead of just once?

lucassshanks: Hmm. I'll think of something later with more thought but if a user is routing to a /login/ route and is already logged in, I'd need to reroute them in that login folders +page.ts file so I'd need to do the auth checks somewhere

cloud/dashboard/src/lib/auth/googleAuth.svelte.ts
lucassshanks: image

lennartkloock: Oh, sorry, nvm then

lucassshanks:
?brawl merge

PR #578, opened 2025-09-15
attempt to split core into multiple sub crates

TroyKomodo:
Attempt to split core into sub crates to improve compiler performance.

### DIFF
--- a/cloud/core/src/common.rs
+++ b/cloud/core/src/common.rs
@@ -246,7 +246,7 @@ pub(crate) async fn create_session<G: core_traits::Global>(
     let session_expires_at = if !mfa_options.is_empty() {
         chrono::Utc::now() + global.timeout_config().mfa
     } else {
-        chrono::Utc::now() + global.timeout_config().user_session_token
+        chrono::Utc::now() + global.timeout_config().user_session
     };
     let token_id = Id::new();
     let token_expires_at = chrono::Utc::now() + global.timeout_config().user_session_token;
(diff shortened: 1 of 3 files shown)


cloud/core/cedar/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+        map.serialize_entry("type", self.type_name.as_str())?;
+        map.serialize_entry("id", self.id.unescaped())?;
+        map.end()
+    }
+}
+
+pub trait CedarIdentifiable {
+    /// MUST be a normalized cedar entity type name.
+    ///
+    /// See [`cedar_policy::EntityTypeName`] and <https://github.com/cedar-policy/rfcs/blob/main/text/0009-disallow-whitespace-in-entityuid.md>.
```
lennartkloock, line 54: Maybe we can remove this comment now?

cloud/core/src/common.rs
```diff
@@ -243,7 +246,7 @@ pub(crate) async fn create_session<G: CoreConfig>(
     let session_expires_at = if !mfa_options.is_empty() {
         chrono::Utc::now() + global.timeout_config().mfa
     } else {
-        chrono::Utc::now() + global.timeout_config().user_session
+        chrono::Utc::now() + global.timeout_config().user_session_token
```
lennartkloock, line 249: This change is wrong. This is the session expiry (30d), not the token expiry (4h).

lennartkloock (changes requested):

lennartkloock (approved):

PR #602, opened 2025-09-26
add diesel cli as a tool

TroyKomodo:
Adds the diesel-cli tool as part of the tools we provide.

This is used `scufflecloud-core` to generate the database schema given migrations.


TroyKomodo:
?brawl merge

PR #603, opened 2025-09-27
fix(core): make mfa credential name optional

lennartkloock:
This PR makes the name field of MFA credentials optional as requested by @lucassshanks


philipch07 (commented):
The code changes seem fine to me but I'm just curious why we would want it to be optional

philipch07 (approved):

PR #606, opened 2025-09-29
remove bazel lock from vcs

TroyKomodo:
The bazel module lock seems to change too often when we make prs, and the changes cannot be automatically resolved with github's default merge resolution making it very annoying to merge prs.


TroyKomodo:
?brawl merge

PR #611, opened 2025-10-03
fix(core): google login for non-google users

lennartkloock:
This PR fixes the Google login for emails that are already registered with us but not associated with a Google account yet. When someone logs in with a Google account which is not linked to a Scuffle user and whose email matches one that is already on record, the Google account will get linked to that existing account now.


lennartkloock:
?brawl merge

